Issue #2, opened 2017-05-30
'append' mode for gzip, bz2 and lzma


pohutukawa:
It looks like Python 3 does support the 'append text' mode through `gzip.open`, `bz2.open` and `lzma.open`. In my projects I'm currently using `gzip.open`, but would like to make it more flexible through the use of `xopen`.

Though, it looks like the integration is done via the `bz2.BZ2File` constructors (and such), which are not supporting the 'append text' mode. Is there a reason for it over using the `.open()` function (maybe for Py 2 compatibility, I didn't check). I'll have a look at finding a way to make it work in a more flexible way.

pohutukawa:
See pull request #3

marcelm:
I’m not sure anymore why I was using `bz2.BZ2File`. I benchmarked everything on multiple Python versions (but that was a few years ago) and sometimes chose non-obvious code to make it a bit faster. However, I just ran some benchmarks again and it does not make a difference. Perhaps `.open()` on Python 3.0/3.1/3.2 does not support the append mode, but I’ve dropped support for these versions anyway, so I’m going with your changes.

pohutukawa:
Maybe in this case then there's some more potential for optimisation/cleaning. I've only changed those occurrences that would give the benefit of the enhanced append mode improvement. Though, I did have the feeling that there were awfully many conditionals involved. Which - in the end - is more common when providing this meta functionality enhancement across different versions. So ... I could imagine further potential for improvement in streamlining/simplifying the code base.

marcelm:
Ok, so the explanation is that [bz2.open() was only added in Python 3.3](https://docs.python.org/3/library/bz2.html#bz2.open). I’ve simplified the code accordingly and made some other small simplifications. Thanks for the tip, and feel free to submit more PRs.

pohutukawa:
It's a shame that the `bz2.BZ2File` constructor and `bz2.open` behave differently with respect to their available modes. But hey, that'd be something to get solved by people on the core CPython team(s) I guess.
Now I reckon we just need a new tag on the code and have it released through PyPI again :-)

Cheers!

pohutukawa:
One last thing: Wouldn't there be the need to replace the reference of `open` in the examples of the `README` with `xopen`?

marcelm:
Of course, thanks! Anything else? Otherwise I’ll make a release.

pohutukawa:
Can't think of anything else but world peace and a halt to global warming right now ... ;-)

[...]

PR #3, opened 2017-05-30
allowing for (text) append mode on bz2 and lzma compression where possible

pohutukawa:
An attempt to solve the problem raised in issue #2. Hopefully the style has been matched sufficiently to make this pull request a smooth one. I've tried it against my locally installed Py 2.7 and 3.5, and it seems to work well. But further testing on other supported platforms should be performed.

### DIFF
--- a/tests/testxopen.py
+++ b/tests/testxopen.py
@@ -18,6 +18,10 @@ try:
 except ImportError:
 	lzma = None
 
+try:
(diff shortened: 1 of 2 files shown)


marcelm:
Great, this looks good! Travis is happy, so I’m merging it. Thanks a lot!

PR #7, opened 2018-04-09
No `sys.stdin.buffer` for Python 2

kristjaneerik:
Ran into an issue when using `xopen.xopen('-', 'r')` in Python 2:

```
>>> xopen.xopen('-', 'r')
AttributeError                            Traceback (most recent call last)
<ipython-input-3-5448057e1aea> in <module>()
----> 1 xopen.xopen('-', 'r')
xopen.pyc in xopen(filename, mode, compresslevel)
    201                         r=sys.stdin,
    202                         rt=sys.stdin,
--> 203                         rb=sys.stdin.buffer,
    204                         w=sys.stdout,
    205                         wt=sys.stdout,

AttributeError: 'file' object has no attribute 'buffer'
```

Fix is to not create the mode lookup dictionary with the `.buffer` options if it's a non-Python 3 environment, which is what the PR does.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ else:
 
 setup(
 	name='xopen',
-	version='0.3.2',
+	version='0.3.3',
 	author='Marcel Martin',
 	author_email='mail@marcelm.net',
 	url='https://github.com/marcelm/xopen/',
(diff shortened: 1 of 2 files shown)


marcelm:
Nice catch, thanks for fixing this!

PR #11, opened 2019-05-16
Open decompression process in gzip subprocess by default.

rhpvorderman:
Stumbled across this library this morning. Wow! Such concise, well-written and readable code. This is one of the best python projects I have ever seen!

I am currently writing a simple fastqsplitter in python that reads a fastq file and splits it into multiple others. This means it does decompression (for the input) and compression (for the outputs) simultaneously.

Unfortunately it is slower than the [scala implementation](https://github.com/biopet/fastqsplitter) because xopen uses `gzip.open` by default when using python 3. This means the business logic of moving the data around and the decompression are done in the same thread. This creates a bottleneck even on my work PC with a core i7-6700 (which should have ample single threaded power). 

With this PR xopen defaults to decompress opened files in separate gzip processes. This increases performance.

before:
```
time ./fastqsplitter.py -i /data/test/U0b_TGACCA_L001_R1.fastq.gz -o split.1.fq.gz -o split.2.fq.gz -o split.3.fq.gz -o split.4.fq.gz -o split.5.fq.gz -c 5

real    1m59.174s
user    8m28.572s
sys     0m12.200s
```

After
```
time ./fastqsplitter.py -i /data/test/U0b_TGACCA_L001_R1.fastq.gz -o split.1.fq.gz -o split.2.fq.gz -o split.3.fq.gz -o split.4.fq.gz -o split.5.fq.gz -c 5

real    1m29.463s
user    8m35.296s
sys     0m15.980s
```
~~Total cpu time has increased, but real time has dropped because of the better parallelization.
This 30 second difference also manifests itself when I use `-c 1`.
It is a trade off, but the 30 seconds extra total cpu time (6% extra compute time) are not as significant compared to the 30 seconds decrease in real time (25% faster finish)~~

EDIT: I made the PipedGzipReader a proper iterator. Now it is just plain faster. No drawbacks. To use the iterator.
```python
myfile = xopen('myfile.txt.gz')
While True:
    try:
        line=next(myfile)
    except StopIteration:
        break
    # Fancy operations with line here.

```

EDIT: You could also use a for loop.

And this is opening one fastq file. If you open two (cutadapt) then it should become even faster in real time.

Hope this helps!

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -205,6 +205,13 @@ class PipedGzipReader(Closing):
         self.process.wait()
         self._raise_if_error()
 
+    def __next__(self):
+        line = self._file.readline()
+        if line:
+            return line
+        else:
+            raise StopIteration()
+
     def _raise_if_error(self):
         """
         Raise IOError if process is not running anymore and the
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -230,6 +230,9 @@ class PipedGzipReader(Closing):
         self._raise_if_error()
         return data
 
+    def readinto(self, *args):
+        data = self._file.readinto(*args)
+        return data
 
 if bz2 is not None:
     class ClosingBZ2File(bz2.BZ2File, Closing):
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -266,8 +266,6 @@ def _open_xz(filename, mode):
 
 
 def _open_gz(filename, mode, compresslevel, threads):
-    if _PY3 and 'r' in mode:
-        return gzip.open(filename, mode)
     if sys.version_info[:2] == (2, 7):
         buffered_reader = io.BufferedReader
         buffered_writer = io.BufferedWriter
@@ -279,7 +277,10 @@ def _open_gz(filename, mode, compresslevel, threads):
             return PipedGzipReader(filename, mode)
         except OSError:
             # gzip not installed
-            return buffered_reader(gzip.open(filename, mode))
+            if _PY3:
+                return gzip.open(filename, mode)
+            else:
+                return buffered_reader(gzip.open(filename, mode))
     else:
         try:
             return PipedGzipWriter(filename, mode, compresslevel, threads=threads)


marcelm:
HI, thanks for the PR! I’ve been wanting to do some benchmarks myself to check how fast `gzip.open` actually is on Python 3. This looks quite good! I’d like to run some tests myself, though (and to make sure that performance in Cutadapt is actually good), but have no time this week, so let me get back to you next week.

rhpvorderman:
> HI, thanks for the PR! I’ve been wanting to do some benchmarks myself to check how fast `gzip.open` actually is on Python 3. This looks quite good! I’d like to run some tests myself, though (and to make sure that performance in Cutadapt is actually good), but have no time this week, so let me get back to you next week.

Thanks for your fast reply. I think gzip.open is quite fast on python3. It uses the same zlib, so there should be no difference. The problem is that it is competing for CPU time with all the other stuff that happens in the python main thread. That is what makes it slow. Especially if you open multiple files.

I am curious what you think next week! Have a nice day!

rhpvorderman:
I added a readinto method so it will work with cutadapt.

rhpvorderman:
Hmm I did some performance benchmarking on cutadapt and  the readinto method of python3 gzip.open is significantly faster than the PipedGzipReader. This means for cutadapt that the old method is much faster.

However for [fastqsplitter](https://github.com/LUMC/fastqsplitter/blob/init/src/fastqsplitter/__init__.py#L85), it is much faster to use the PipedGzipReader. It finishes faster in wall clock time, with no cost in total cpu time. 

Splitting fastq into two fastqs with gzip.open:
```
real	1m45.809s
user	3m9.948s
sys	0m5.860s
```
Splitting fastq into two fastqs with pipedgzipreader:
```
real	1m12.826s
user	3m10.792s
sys	0m6.824s
```
Possibly cutadapt can be optimized to make use of the extra parallelization, but for now it is not good to change the default behaviour.

I can solve this by adding a boolean variable to xopens input called `prefer_native_open=False`. If set to true this will prefer to open in piped(gzip|xz|bzip2)reader else it will default to the python method for opening these.

@marcelm What is your opinion on this?

marcelm:
Very interesting.

I’ve done some time measurements myself and can confirm the following observations:

* `pigz` is a lot faster than `gzip` when *de*compressing files (three times faster in my tests). This is surprising as the man page is quite cautious about decompression performance:

> Decompression can't be parallelized, at least  not  without  specially  prepared
> deflate  streams  for that purpose.  As a result, pigz uses a single thread (the
> main thread) for decompression, but will create three other threads for reading,
> writing, and check calculation, which can speed up decompression under some cir‐
> cumstances.

* `PipedGzipReader.read()` is a lot slower than `gzip.open().read()`.

I’ll try to run a profiler and see what comes up.

Here’s how to reproduce the results (run from master branch in an xopen checkout).

----

## Setup

Install Pythons

    for v in 2.7 3.4 3.5 3.6 3.7; do conda create -y -p py$v python=$v; py$v/bin/pip install --upgrade pip; py$v/bin/pip install -e .; done

Create a 100 MB test file

    head -c 100000000 /dev/urandom > data
    gzip < data > data.gz


## Decompression

### With gzip

    time gzip -dc < data.gz > /dev/null

real    0m0.477s
user    0m0.465s
sys     0m0.012s

### With pigz

    time pigz -dc < data.gz > /dev/null

real    0m0.123s
user    0m0.106s
sys     0m0.074s

### With gzip.open

    for v in 2.7 3.4 3.5 3.6 3.7; do printf "\nPython %s: " $v; py$v/bin/python -m timeit -n 4 -r 3 -s "import gzip" "gzip.open('data.gz', 'rb').read()"; done

Python 2.7: 4 loops, best of 3: 378 msec per loop
Python 3.4: 4 loops, best of 3: 384 msec per loop
Python 3.5: 4 loops, best of 3: 257 msec per loop
Python 3.6: 4 loops, best of 3: 257 msec per loop
Python 3.7: 4 loops, best of 3: 259 msec per loop

### With PipedGzipReader

    for v in 2.7 3.4 3.5 3.6 3.7; do printf "\nPython %s: " $v; py$v/bin/python -m timeit -n 4 -r 3 -s "from xopen import PipedGzipReader" "PipedGzipReader('data.gz', 'rb').read()"; done

Python 2.7: 4 loops, best of 3: 561 msec per loop
Python 3.4: 4 loops, best of 3: 557 msec per loop
Python 3.5: 4 loops, best of 3: 557 msec per loop
Python 3.6: 4 loops, best of 3: 555 msec per loop
Python 3.7: 4 loops, best of 3: 557 msec per loop

marcelm:
Whoops, ok, the `PipedGzipReader` always uses `gzip`, never `pigz`, so then it’s obvious that it cannot be faster than `gzip.open`. I would expect some improvements because it runs in a background process, though.

marcelm:
I think it makes sense to take your commits up to now in reverse order, so I’ve cherry-picked the one about adding `PipedGzipReader.readinto`. That makes sense no matter how we proceed. I’ve also added unit tests.

rhpvorderman:
Thanks. I will do some benchmarking as well with the new code as well and see if there are differences and possible improvements.  I will also resolve the merge conflicts. It would be nice if pipedgzipreader was always the fastest and xopen could default to that. That would simplify the decision tree.

> Create a 100 MB test file

May I note that random data does not compress very well? Compression looks at patterns in the data. In random data there should be no patterns. So I don't know how representative this benchmark is.

I benchmarked on these files:  ftp://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/data/NA12878/NIST_NA12878_HG001_HiSeq_300x/131219_D00360_005_BH814YADXX/Project_RM8398/Sample_U0b/

marcelm:
You’re right about the random data. I’m aware that it doesn’t compress, but was assuming that the relative differences in performance would be the same. Here are some updated benchmarks for decompression.

## Gzipped FASTQ file, low compression level

This is the file `U0b_TGACCA_L001_R1_001.fastq.gz` from the FTP directory you linked to above.

Method | Time (seconds)
-|-
gzip | 14.4
pigz | 6.2 (7.7 CPU)
gzip.open on Python 2.7 and 3.4 | 23.9
gzip.open on Python 3.5+ | 9.5
PipedGzipReader with pipe to gzip | 14.4
PipedGzipReader with pipe to pigz | 6.7

## Gzipped FASTQ file, compression level 6

This is the same FASTQ file as above, but recompressed to compression level 6. (It shrinks from 480 MB to 410 MB.)

Method | Time (seconds)
-|-
gzip (command-line) | 10.2
pigz (command-line) | 5.7 (7.2 CPU) 
gzip.open on Python 2.7 and 3.4 | 21.5
gzip.open on Python 3.5+ | 8.8
PipedGzipReader with pipe to gzip | 11.3
PipedGzipReader with pipe to pigz | 6.1

So even for real data, the `PipedGzipReader` with a pipe to pigz is the best although the difference is not as big as for random data.

Would be great to see what you find in your benchmarks.

marcelm:
There’s one commit by you "implement proper iterator" that I haven’t merged, yet. I’m pretty sure it is not necessary (`__iter__` is already implemented and returns a generator so implementing `__next__` is redundant). Are you sure you saw any performance improvements from that?

rhpvorderman:
From the python wiki https://wiki.python.org/moin/Iterator:

> An iterable object is an object that implements __iter__, which is expected to return an iterator object.

> An iterator is an object that implements next, which is expected to return the next element of the iterable object that returned it, and raise a StopIteration exception when no more elements are available. 

A proper iterator also implements a next method. Only `__iter__` is not enough. (`__iter__` is necessary for any list-like object that you can iterate over).

Yes, there are significant performance gains when using.
```python
xopen("file.gz") as gzip_file:
     while True:
          my_function(next(gzip_file))  # As opposed to my_function(next(iter(gzip_file)))
```

rhpvorderman:
> Would be great to see what you find in your benchmarks.

Wow. Thanks for doing all this great benchmarking stuff. I will start testing right away and post the results.

marcelm:
> A proper iterator also implements a next method.

The generator returned by `__iter__()` implicitly implements `__next__`.

Yes, the other choice is to have `__iter__()` return `self` and implement `__next__` as well. Then `PipedGzipReader` would be both the iterable and its own iterator. For normal container objects, this would be a bad design choice because you could then only iterate once over them. It works in this case because file-like objects are expected have this behavior.

Regarding your example code, I don’t quite get why you write it like this. The “natural” way to use an iterable is with a `for` loop. When you write `for item in items: ...`, the `items` needs to be an iterable, and an iterator will be obtained for it only once at the beginning. I think it is equivalent to this:
```python
it = iter(items)
while True:
  try:
    item = next(it)
  except StopIteration:
    break
  ...
```
You wrote something like this in your very first post above, but note the additional first line here, which makes it clear that `__iter__` is only called once.

marcelm:
Regarding the potential speedup in Cutadapt: Perhaps you’ve seen that I’ve moved FASTQ parsing to a separate library called `dnaio`, which Cutadapt now uses. `dnaio` avoids using `__iter__`. Instead, it reads in large chunks of data and looks for the positions of line breaks manually (in Cython). It’s a lot faster this way.

marcelm:
The changes this PR was originally about have been merged, so I’d like to close this now. Please feel free to continue work on the `__iter__` thing and then open a new PR.

Since I’ve done the merging manually and would like to reject the remaining changes, I can only mark this as "Closed". If you would like to see a nice "Merged", remove your `__iter__` changes and add a line to the credits and then I can merge. But this is just for GitHub esthetics. If you don’t, I’ll add the credits before the next release.

rhpvorderman:
@marcelm Thanks. I benchmarked a bit and the `__iter__` method is consistently faster than the `__next__` method (by5-10% or so ) so this can be closed. I indeed did implement it in the wrong way, as it created an iterator object on each invocation which is extremely wasteful. Implemented in the correct way it is faster than `__next__`.

I use a while loop and not a for loop, because it makes it somewhat easier algorithmically to get groups of 100 reads and put them into different output files. 

Now that the pipedgzipreader is default, this will make fastqsplitter a lot faster. I will look into dnaio and see if it can make fastqsplitter even faster. Thanks for all the benchmarks and feedback, it was a lot of fun doing this PR.

P.S. Any plans on making a new release soon?

marcelm:
> Thanks for all the benchmarks and feedback, it was a lot of fun doing this PR.

Yes, it was, and thanks again for your contribution! I don’t get that many PRs on my projects, so I’m usually happy when I do.

> Any plans on making a new release soon?

I will do this right now. Since I automated this a while ago, I will only need to tag a commit and push it.

rhpvorderman:
Great!!
As a bonus (you seem to like performance tables as much as I do :wink:) here is the speed diff of fastqsplitter. Before (denoted with -): using gzip for decompression and wrong implementation of next. After (denoted with+) using pigz and proper implementation of iter. Your feedback has improved the application considerably (saving 26 cpu seconds on default!). So lots of thanks for that.

```
 |                       | fastqsplitter (defaults) | fastqsplitter -t 4 -c 5 |  biopet-fastqsplitter |
 | --------------------  | ------------------------ | ----------------------- | --------------------- |
-|real time              | 0m57.632s                | 1m30.291s               | 1m41.385s             |
-|total cpu time         | 3m33.276s                | 8m36.600s               | 8m20.304s             |
+|real time              | 0m50.932s                | 1m28.153s               | 1m41.385s             |
+|total cpu time         | 3m7.116s                 | 7m55.436s               | 8m20.304s             |
 |max mem                | 24 MB                    | 32MB                    | 400MB                 |
-|max vmem               | 82 MB                    | 1.6 GB                  | 11.0 GB               |
+|max vmem               | 110 MB                   | 1.6 GB                  | 11.0 GB               |
 |output files total size| 2290 MB                  | 2025 MB                 | 2025 MB               |
```

PR #1, opened 2016-09-15
Fix github link

kyleabeauchamp:
I updated my contact link to use the account I typically use for my open source work.

PR #14, opened 2019-08-12
Improve speed for pipedgzip reader.

rhpvorderman:
Hi, I have been doing some major performance nerding lately as part of my work on fastqsplitter.

I discovered that
```
with open('my_file.txt', 'rb') as file_handler:
     for line in file_handler:
        # do stuff per line
        pass
```
Is the fastest way to iterate over files in python. `file_handle.readline()` or `next(file_handle)` exhibits some overhead.
I noticed that the PipedGzipReader uses a lot of cpu time when iterating over it that way. This PR fixes that by letting `__iter__` return `self._file` as that is already an iterator.

I did some benchmarks with this script:
```
#/usr/bin/env python3

import sys
import xopen

with xopen.xopen(sys.argv[1]) as file_h:
    for line in file_h:
        pass
```

I conducted some testing with [hyperfine](https://github.com/sharkdp/hyperfine) (installable via conda). It is quite a nice tool. It does several runs and allows you to set a warmup so the file is thoroughly cached in memory before the actual tests start.

Before:
```
hyperfine -w 3 "python3 test.py ~/test/big.fastq.gz"
Benchmark #1: python3 test.py ~/test/big.fastq.gz
  Time (mean ± σ):      8.263 s ±  0.239 s    [User: 15.095 s, System: 1.841 s]
  Range (min … max):    7.982 s …  8.657 s    10 runs
 ```

after:
```
hyperfine -w 3 "python3 test.py ~/test/big.fastq.gz"
Benchmark #1: python3 test.py ~/test/big.fastq.gz
  Time (mean ± σ):      5.847 s ±  0.177 s    [User: 12.541 s, System: 1.801 s]
  Range (min … max):    5.567 s …  6.212 s    10 runs
```

Xopen still crashes when you give it a corrupt gzip file. The output is the same. All the tests succeed. The speedup is quite nice: >25%. And the resulting code is smaller.

What more do you need? :wink:

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -215,10 +215,7 @@ class PipedGzipReader(Closing):
         self._raise_if_error()
 
     def __iter__(self):
-        for line in self._file:
-            yield line
-        self.process.wait()
-        self._raise_if_error()
+        return self._file
 
     def _raise_if_error(self):
         """

### CODE REVIEW

**Comment by rhpvorderman** [comment]
For context I did a hyperfine run of `pigz -cd` itself:
```
hyperfine -w 3 "pigz -cd ~/test/big.fastq.gz >> /dev/null"
Benchmark #1: pigz -cd ~/test/big.fastq.gz >> /dev/null
  Time (mean ± σ):      4.310 s ±  0.097 s    [User: 6.783 s, System: 0.441 s]
  Range (min … max):    4.144 s …  4.516 s    10 runs
 ```
Using pigz alone it uses 160% cpu or 1.6 core, therefore the real time is less than the cpu time.
pigz takes on average 6.8 cpu seconds to decompress this file.
Before this PR: 15.1 - 6.8  = 8.3 cpu seconds python overhead
After this PR: 12.6 - 6.8 = 5.8 cpu seconds python overhead.
5.8 / 8.3 ~= 0.7. A 30% reduction in python overhead is achieved.

Given that the python CPU overhead equals the real time it shows that python is the limiting factor still (I can also see this on my system monitor, python reaches 100% of one core all the time). `pigz` decompresses faster than python can iterate over lines.

marcelm:
Thanks, this is interesting!

I had to do some digging why the original code was written this way. After all, your patch removes the `self.process.wait()` and `self._raise_if_error()` lines, which are obviously meant for error handling. This is the original commit in Cutadapt: https://github.com/marcelm/cutadapt/commit/63b787617e1564e4838b3b4692789f0ddfd3b523

That commit introduces the `PipedGzipReader` class. Previously, I just used `subprocess.Popen(...)` directly, but this hid errors when reading truncated gzip files.

Interestingly, all tests pass with your patch, in particular even those that check reading from truncated gzip files raises an error!

Here is one of those tests (abbreviated):
```
f = xopen("truncated.gz", "r")
for line in f:
    pass
f.close()
```
As it turns out, the current version will correctly crash on one of the `for line in f` iterations, and the changed version will not do so. However, your version will crash on `f.close()`, so the test still passes.

So this change will be an inconvenience for those users who do not close their input files, but one can argue that it’s already a bug in their code.

One alternative is to use a `yield from self._file` instead, and then the error checking can be left in. Do you think you could run your time measurements for that as well? If the runtime is the same, I’d prefer this, but otherwise, I’m probably fine with merging it as-is.

rhpvorderman:
I realized that no errors would be checked in the iter method, but since the checking self._raise_if_error() was also done in the close() method I figured that the checking would be done anyway. If files are not properly closed that leads to other problems as well.

I tested the following snippet:

```python
    def __iter__(self):
        yield from self._file
        self.process.wait()
        self._raise_if_error()
```
Benchmark:
```
hyperfine -w 3 "python3 test.py ~/test/big.fastq.gz"
Benchmark #1: python3 test.py ~/test/big.fastq.gz
  Time (mean ± σ):      7.727 s ±  0.265 s    [User: 14.558 s, System: 1.881 s]
  Range (min … max):    7.257 s …  8.138 s    10 runs
```
So that is considerably slower than returning self._file.

marcelm:
Thanks for running the extra benchmark! This is a significant improvement, thanks a lot.

rhpvorderman:
@marcelm, could you make a new release? It is nice to have the extra speed available on pypi. Thanks!

marcelm:
Sure, I just tagged 0.8.0.

However, I’m a bit worried that something is now broken because [of this test failure](https://travis-ci.org/marcelm/xopen/builds/571788800). This is the test run for the commit where I only updated the changelog, so it should pass. The next build does indeed pass, which is why the automatic release to PyPI went through. Possibly this is a non-deterministic failure.

marcelm:
I don’t have time at the moment, but have opened #16 to keep track of this.

PR #20, opened 2019-10-18
limit the number of threads also for reading

bernt-matthias:
using the same logic as for writing

this is really important for me (assuming that it works :) ) since cutadapt can not be used on restrictive cluster environments otherwise (on out cluster jobs are killed if they take more resources than requested -- and I can't request all CPUs).

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -198,9 +198,11 @@ class PipedGzipReader(Closing):
             raise ValueError("Mode is '{0}', but it must be 'r', 'rt' or 'rb'".format(mode))
         if threads is None:
             threads = min(_available_cpu_count(), 4)
+        
+        pigz_args = ['pigz', '-cd', path]
         if threads != 0:
             pigz_args += ['-p', str(threads)]
-        self.process = Popen(['pigz', '-cd', path] + pigz_args, stdout=PIPE, stderr=PIPE)
+        self.process = Popen(pigz_args, stdout=PIPE, stderr=PIPE)
         self.name = path
         if _PY3 and 'b' not in mode:
             self._file = io.TextIOWrapper(self.process.stdout)
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -190,13 +190,17 @@ class PipedGzipReader(Closing):
     also faster than gzip when reading (three times faster).
     """
 
-    def __init__(self, path, mode='r'):
+    def __init__(self, path, mode='r', threads=None):
         """
         Raise an OSError when pigz could not be found.
         """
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Hope that this fixes https://github.com/marcelm/cutadapt/issues/290 .. I could help there.

marcelm:
Thanks, this looks good, and I agree, this is an important fix! It’s definitely an oversight that this was missing.

Have you tested whether this fixes the problem on your cluster?

marcelm:
Version 0.8.3 is now on PyPI.

bernt-matthias:
Will test as soon as possible.

rhpvorderman:
> Hope that this fixes marcelm/cutadapt#290 .. I could help there.

It won't. I am sorry to say this.

From the pigz manual: 

> Decompression  can’t  be  parallelized,  at  least  not  without  specially  prepared  deflate  streams  for that purpose. As a result, pigz uses a single thread (the main thread) for decompression, but will create three other threads for reading, writing, and check calculation, which can speed up decompression under some circumstances. Parallel decompression can be turned off by specifying one process (-dp 1 or -tp 1).

This means pigz will use at most 4 threads.
The python code:

```python
        if threads is None:
            threads = min(_available_cpu_count(), 4)
```

Limits pigz to a default of 4 threads. Which is already the max amount used when decompressing. So it is redundant and can be removed.

The rest of the code is good and quite nice as it allows developers to override the default of `None` and use `1` instead to disable parallel decompression. There is no flag in cutadapt for this as far as I am aware. If such a flag were to be provided in conjunction with this code I think that would help towards solving the issue.

@bernt-matthias  I have done a lot of benchmarking with pigz and from my experience the decompression only takes 160% of cpu (1.6 core) when decompressing. This is the max. The exorbitant use of threads is only caused by pigz compressing. An easy solution for this is to set the compression level to 1. (Shorthand flag `-Z` in cutadapt). This will drastically lower the cpu requirements.

marcelm:
@rhpvorderman Ok, sorry for merging so eagerly. I’ll wait for a review next time.

@bernt-matthias Can you share some more details about the cluster system? I’d like to know how it detects that a job is using too much CPU.

bernt-matthias:
@marcelm I don't know the exact details. But I guess the admins have some monitoring system triggering an alarm when a load > 100% on a compute node is detected. 

Our admins are very strict regarding fair usage of resources. Which makes sense to me, since as a user I can be sure that the requested resources are available to my processes.

> An easy solution for this is to set the compression level to 1. (Shorthand flag -Z in cutadapt). This will drastically lower the cpu requirements.

Thanks for the suggestion.

PR #21, opened 2019-10-21
Limit thread usage to 1 for decompressing.

rhpvorderman:
Pigz does not use more than 4 threads when decompressing. Therefore defaulting to 4 threads is not necessary.

This is more of a technical debt thing, and does not warrant a new release. Still every line removed is a line that does not need to be maintained.

By the way, should `def _available_cpu_count():` still be there? Is there a use case that `multiprocessing.cpu_count()` does not cover? (Maybe python2, but does that need to be supported anymore? Cutadapt does not support python2 anymore.)

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -198,16 +198,17 @@ class PipedGzipReader(Closing):
             raise ValueError("Mode is '{0}', but it must be 'r', 'rt' or 'rb'".format(mode))
 
         pigz_args = ['pigz', '-cd', path]
-        if threads:  # Check if threads is not 0 or None.
-            pigz_args += ['-p', str(threads)]
-        else:
+
+        if threads is None or threads == 0:
             # Single threaded behaviour by default because:
             # - Using a single thread to read a file is the least unexpected
             #   behaviour. (For users of xopen, who do not know which backend is used.)
-            # - There is quite a substantial overhead (+25% cpu time) when
+            # - There is quite a substantial overhead (+25% CPU time) when
             #   using multiple threads while there is only a 10% gain in wall
-            #   clock time. This is a bad trade off.
-            pigz_args += ['-p', '1']
+            #   clock time.
+            threads = 1
+
+        pigz_args += ['-p', str(threads)]
 
         self.process = Popen(pigz_args, stdout=PIPE, stderr=PIPE)
         self.name = path
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -198,7 +198,7 @@ class PipedGzipReader(Closing):
             raise ValueError("Mode is '{0}', but it must be 'r', 'rt' or 'rb'".format(mode))
 
         pigz_args = ['pigz', '-cd', path]
-        if threads != 0:
+        if threads != 0 and threads is not None:
             pigz_args += ['-p', str(threads)]
         self.process = Popen(pigz_args, stdout=PIPE, stderr=PIPE)
         self.name = path
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -198,7 +198,7 @@ class PipedGzipReader(Closing):
             raise ValueError("Mode is '{0}', but it must be 'r', 'rt' or 'rb'".format(mode))
 
         pigz_args = ['pigz', '-cd', path]
-        if threads != 0 and threads is not None:
+        if threads:  # Check if threads is not 0 or None.
             pigz_args += ['-p', str(threads)]
         self.process = Popen(pigz_args, stdout=PIPE, stderr=PIPE)
         self.name = path
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -199,7 +199,7 @@ class PipedGzipReader(Closing):
 
         pigz_args = ['pigz', '-cd', path]
 
-        if threads is None or threads == 0:
+        if not threads:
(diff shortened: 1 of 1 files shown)


marcelm:
The tests fail, but I agree in principle with this change, feel free to merge as soon as the tests pass.

`_available_cpu_count` needs to be there as long as we support Python 2.7, which is currently the case. I’d like to remove support for it when Python 2.7 is retired on January 1st, 2020. Then also some other code can go.

marcelm:
It seems pigz in decompression mode has two settings: Use threads or not, so perhaps the `threads` parameter should actually be a boolean for choosing between these options. Then it should be named `use_threads` so it’s clearer what it does.

It appears that `pigz` is faster than `gzip` when decompressing even with `-p 1`, so `pigz` should be preferred over `gzip` even in that case.

marcelm:
> I agree, but this will be hidden at the xopen interface, that implements threads as a number.

You’re right, I forgot, so passing it through is fine.

src/xopen/__init__.py
```diff
@@ -196,11 +196,9 @@ def __init__(self, path, mode='r', threads=None):
         """
         if mode not in ('r', 'rt', 'rb'):
             raise ValueError("Mode is '{0}', but it must be 'r', 'rt' or 'rb'".format(mode))
-        if threads is None:
-            threads = min(_available_cpu_count(), 4)
-        
+
         pigz_args = ['pigz', '-cd', path]
-        if threads != 0:
+        if threads != 0 and threads is not None:
```
marcelm: This should probably be simplified to `if threads:`

rhpvorderman:
@marcelm added two commits. One to change the conditional to something better.
The other one reflects the discussion in https://github.com/marcelm/cutadapt/pull/412 . Decompression threads should be 1. This means behaviour will be the same for opening `.xz` `.bz2` or `.gz` files. This will be the least unexpected behaviour, while still making use of the massively better decompression pigz provides over gzip.

src/xopen/__init__.py
```diff
@@ -196,12 +196,19 @@ def __init__(self, path, mode='r', threads=None):
         """
         if mode not in ('r', 'rt', 'rb'):
             raise ValueError("Mode is '{0}', but it must be 'r', 'rt' or 'rb'".format(mode))
-        if threads is None:
-            threads = min(_available_cpu_count(), 4)
-        
+
         pigz_args = ['pigz', '-cd', path]
-        if threads != 0:
+        if threads:  # Check if threads is not 0 or None.
```
marcelm: This is a redundant comment. It should be removed.

src/xopen/__init__.py
```diff
@@ -196,12 +196,19 @@ def __init__(self, path, mode='r', threads=None):
...
             pigz_args += ['-p', str(threads)]
+        else:
+            # Single threaded behaviour by default because:
+            # - Using a single thread to read a file is the least unexpected
+            #   behaviour. (For users of xopen, who do not know which backend is used.)
+            # - There is quite a substantial overhead (+25% cpu time) when
+            #   using multiple threads while there is only a 10% gain in wall
+            #   clock time. This is a bad trade off.
```
marcelm: Nitpicking: Please remove "This is a bad trade off." (The reader can decide that for themselves.) Please write uppercase "CPU".

src/xopen/__init__.py
```diff
@@ -196,12 +196,19 @@ def __init__(self, path, mode='r', threads=None):
         """
         if mode not in ('r', 'rt', 'rb'):
             raise ValueError("Mode is '{0}', but it must be 'r', 'rt' or 'rb'".format(mode))
-        if threads is None:
-            threads = min(_available_cpu_count(), 4)
-        
+
         pigz_args = ['pigz', '-cd', path]
```
marcelm: You can remove some duplication by writing it like this:
```
if threads is None:
    # (Put the comment from below here)
    threads = 1
pigz_args = ['pigz', '-p', str(threads), '-cd', path]
```

src/xopen/__init__.py
rhpvorderman: I extremely dislike the implicitness of truthy/falsey testing. In this case the possible value is an optional integer that may be `0` or `None`. I do not want to test for `[]`, `{}`, `()`, `False` or empty set. Hence the comment to clarify what values I am expecting. 
But this is a style thing and just my personal taste, so I will remove it.

marcelm: Thanks.

src/xopen/__init__.py
```diff
@@ -196,12 +196,20 @@ def __init__(self, path, mode='r', threads=None):
...
-        if threads != 0:
-            pigz_args += ['-p', str(threads)]
+
+        if not threads:
```
marcelm: Hm, now I would actually write it as `threads is None` because the purpose is to explicitly check whether the user set something different than the default `None`, not to check the falsiness. Perhaps I’m being inconsistent here, but since you found the more explicit version more readable anyway, can you change it? Sorry for the back and forth.

marcelm: Since it’s me who’s to blame here, I’ll just merge this and fix it up afterwards.

rhpvorderman: Well it actually was `if threads is None or threads == 0:`
But then I thought I was droning on again about my style preference.

`0` is a weird case anyway. Maybe that should be reserved for running the pigz default?

PR #23, opened 2019-10-23
Allow threads=0 to mean that no external pigz/gzip should be run

marcelm:
Instead, regular gzip.open() is used.

I’m planning to use this to help solving marcelm/cutadapt#290.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -335,25 +335,25 @@ def _open_gz(filename, mode, compresslevel, threads):
         exc = OSError
 
     if 'r' in mode:
-        def with_threads():
+        def open_with_threads():
             return PipedGzipReader(filename, mode, threads=threads)
 
-        def without_threads():
+        def open_without_threads():
             return buffered_reader(gzip.open(filename, mode))
     else:
-        def with_threads():
+        def open_with_threads():
             return PipedGzipWriter(filename, mode, compresslevel, threads=threads)
 
-        def without_threads():
+        def open_without_threads():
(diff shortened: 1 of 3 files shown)


marcelm:
I wondere whether the interface is good enough. Setting `threads=0` is sometimes used for using all available cores. The alternative is to use an extra parameter, something like `subprocess=False`.

rhpvorderman:
> I wonder whether the interface is good enough. Setting threads=0 is sometimes used for using all available cores. The alternative is to use an extra parameter, something like subprocess=False.

Well since `threads=0` spawns 0 extra threads, it is quite good.
I think threads=None is okay for all available cores or other sane default.

src/xopen/__init__.py
```diff
@@ -333,18 +333,28 @@ def _open_gz(filename, mode, compresslevel, threads):
...
+        def with_threads():
             return PipedGzipReader(filename, mode, threads=threads)
-        except exc:
-            # pigz is not installed
+
+        def without_threads():
             return buffered_reader(gzip.open(filename, mode))
     else:
-        try:
+        def with_threads():
```
rhpvorderman: These functions obscure what is happening and diminish readability. It seems better to me to use the return statements directly in the `if threads==0` block.

marcelm: I want to avoid code duplication: The `except` and the `threads==0` cases are the same. The only way I can see to not have to duplicate the code is to move it into a function.

marcelm: Even after considering this for a while, I think this is more readable than the alternative (also because there’s now only a single try/except). I’ve changed the function names a bit, however, to improve readability.

PR #25, opened 2019-10-28
Fix linting errors

rhpvorderman:
The latest master commit failed due to lint errors. Probably because the linting PR and the threads functionality PR were made simultaneously. Hence the threads functionality was not checked against the linter before being merged.

Apparantly the linter thought _open_gz was too complex. Probably because of the function definitions. Here is my take on it. I think it is a bit easier to understand and it is also less complex (according to the linter). What do you think?


marcelm:
That is simpler, I agree.

Issue #4, opened 2017-06-01
Idea: support for file-type detection via magic numbers


winni2k:
I just came across this project, and I was wondering if there was any particular reason that the file type is being inferred from the file name ending instead of the file's [magic number](http://www.astro.keele.ac.uk/oldusers/rno/Computing/File_magic.html)?

marcelm:
The main reason is that I started with name-based detection and it has worked well enough so far. Another reason is that named pipes aren’t seekable, which complicates the implementation if we want the content-based detection to work also for them.

wkretzsch:
Fair enough. 

Aren't named pipes peek-able?

marcelm:
This has now been implemented thanks to a contribution from @bvaisvil . I’m releasing version 0.9.0 at the moment, which includes this change.

PR #28, opened 2020-01-10
Detecting compressed file formats with byte checks

bvaisvil:
This solves a problem for me where my files do not end in ".gz" or ".xz" but are compressed FASTQ files. 

If the filename doesn't end in ".gz" or ".gz" or ".bz2" 6 bytes are read and tested for standard file signatures.

- gz : 0x1f 0x8b (https://tools.ietf.org/html/rfc1952#page-6)
- bz2: 0x42 0x5A 0x68 (https://en.wikipedia.org/wiki/List_of_file_signatures)
- xz : 0xfd 0x37 0x7a 0x58 0x5a 0x00 (https://tukaani.org/xz/xz-file-format.txt)

I've added three additional tests and files.


bvaisvil:
Looks like the new tests failed in python2.7. I'll update the PR with a conditional not to run the tests for python2.

marcelm:
Cool! I’ll give it a closer look next week. Some comments already now:

- This would close #4 
- We’re actually planning to remove Python 2.7 support, so you could ignore tests failing on 2.7.
- You can simplify the byte comparisons. Instead of testing each byte individually, just check whether `bs[0:2] == b'BZh'` in the case of bzip2, for example.

bvaisvil:
- Great. Actually, I had forgot to check the issues on this repository (oops).
- Okay. However, I was able to make 2 of the 3 formats work for python 2.7.
- That would work better and also avoid processor endianness issues. I'll make that change.

tests/test_xopen.py
```diff
@@ -117,6 +117,14 @@ def test_pipedgzipreader_textiowrapper():
             wrapped = io.TextIOWrapper(f)
             assert wrapped.read() == CONTENT
 
+def test_bytetested_gzip():
```
marcelm: Regarding the added `tests/file.txt....test` files (which I cannot comment on directly): Can you please add them as symbolic links? That would make it clearer that they are actually identical to the versions without the `.test` suffix.

marcelm: How about `test_detect_gzip_file_format_from_content` as function name? It’s fine if test function names are long and this would express clearer what the test does. Same for the other two tests.

src/xopen/__init__.py
```diff
@@ -402,6 +402,24 @@ def xopen(filename, mode='r', compresslevel=6, threads=None):
         return _open_xz(filename, mode)
     elif filename.endswith('.gz'):
         return _open_gz(filename, mode, compresslevel, threads)
+    elif mode.startswith("r"):
+        # Test up to the first 6 bytes to detect the file format.
+        with open(filename, "rb") as fh:
```
marcelm: There are some circumstances under which this can fail, for example, if the given filename is `/dev/stdin` or `/dev/fd/0`. That would currently work as expected because there’s only one call to `open()`. However, with this change, there would be two calls and then the second would fail.

Can you please change it such that the file is only opened and read for content detection when it is a regular file? You could check this with `stat.S_ISREG(os.stat(filename).st_mode)`.

Another [solution for a similar file-detection problem, which I implemented in another project](https://github.com/marcelm/dnaio/blob/24f94975f2ac09f682a5486cc08fda4e6727fbcc/src/dnaio/__init__.py#L197) is to open the file just once, read from it, seek back to the beginning and then pass on the already open file descriptor (or use `peek()`). Here, this would imply that, for example, `gzip.open` would need to get an open, file-like object (this is possible with `gzip.open(fileobj=...)`), but then when you call `.close()` on the returned GzipFile, the underlying file descriptor is not closed. So while this method would even allow you to correctly open gzip-compressed standard input, there would be some bigger changes involved.

src/xopen/__init__.py
```diff
@@ -402,6 +402,24 @@ def xopen(filename, mode='r', compresslevel=6, threads=None):
...
+            bs = fh.read(6)
+            if not _PY3:
+                bs = bytearray(bs)
+        if bs[:2] == b'\x1f\x8b':
+            # https://tools.ietf.org/html/rfc1952#page-6
+            return _open_gz(filename, mode, compresslevel, threads)
```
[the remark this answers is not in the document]
marcelm: Ok, perhaps it’s worth changing this since `flake8` complains abotu `xopen()` becoming too complex. Moving the content-based file type detection into a separate function would take care of this.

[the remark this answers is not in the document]
bvaisvil: I ended up refactoring the name detection and content detection into two difference functions. However, `flake8` still thinks the function is too complex. I'm open to other suggestions...

tests/test_xopen.py
bvaisvil: Changed in latest commit

src/xopen/__init__.py
bvaisvil: This is a good point I hadn't considered. I added the check you suggested: `stat.S_ISREG(os.stat(filename).st_mode)`. However, if the file doesn't exist that returns an `OSError`. At the moment I change this to an `IOError` but not sure that is the best approach.

src/xopen/__init__.py
```diff
@@ -356,6 +356,46 @@ def _open_gz(filename, mode, compresslevel, threads):
         return buffered_writer(gzip.open(filename, mode, compresslevel=compresslevel))
 
 
+def _detect_format_from_content(filename, mode, compresslevel, threads):
```
marcelm: It’s great you factored out this function, but what I initially thought was to separate detecting file contents from opening the file (I should have been clearer about this, sorry): So `_detect_format_from_content` and `_detect_format_from_name` would actually only do the detection work and return the file type (perhaps as a string `bz2`, `xz`, `gz` or None), and then within the `xopen` function itself, this result gets used to use the correct function to open the file. As it is now, the `_detect` functions also do the opening, so I’d say that at least their names aren’t quite accurate.

And regarding the exception, I think it would be fine to catch the `IOError` and just return None as "detection result" (for "unknown"). Detecting the format from the name works like this as well: If the name isn’t recognized, it just falls through to a regular `open()`.

tests/file.txt.bz2.test
```diff
@@ -0,0 +1 @@
+file.txt.bz2
```
rhpvorderman: This file is not binary (according to github). So you are not actually testing the opening of binary files.

tests/test_xopen.py
```diff
@@ -118,6 +118,16 @@ def test_pipedgzipreader_textiowrapper():
             assert wrapped.read() == CONTENT
 
 
+def test_detect_gzip_file_format_from_content():
+    with xopen("tests/file.txt.gz.test", "rb") as fh:
```
rhpvorderman: Why are the tests opening plaintext files? That will always succeed, but not touch the code that you wrote.

tests/file.txt.bz2.test
marcelm: I think this is just GitHub not displaying the symlink correctly.

But it would actually be an idea not to add any files at all, but to instead just create a temporary copy the existing file as part of the test. Then the repo itself isn’t cluttered with the symlinks.

rhpvorderman: Ah thanks! But why create temporary files? These test files can be opened as is, right? As far as I can see the tests do nothing special (with regards to opening).

marcelm: To test whether the file format is recognized even though the file name extension is missing, we need (among others) a gzip file whose name does not end in `.gz`. We can get that the way it’s done currently: Add an extra file or symbolic link that has the appropriate name to the tests/ folder permanently. What I’m suggesting is that we could instead create the file each time the test is run from one of the existing files in the tests/ folder.

Or are we talking about different things?

src/xopen/__init__.py
bvaisvil: Oh, I think you communicated what you wanted perfectly. I was trying to be clever and avoid opening the file, if possible, while keeping the number of conditionals to a minimum. 

But, would something along the lines the code below work better?
```python
 if filename == '-':
     return _open_stdin_or_out(mode)
detected_format = _detect_format_from_content(filename, mode)
if filename.endswith('.bz2') or detected_format == "bz2":
     return _open_bz2(filename, mode)
elif filename.endswith('.xz') or detected_format == "xz":
     return _open_xz(filename, mode)
elif filename.endswith('.gz') or detected_format == "gz":
     return _open_gz(filename, mode, compresslevel, threads)
else:
     # Python 2.6 and 2.7 have io.open, which we could use to make the returned
     # object consistent with the one returned in Python 3, but reading a file
     # with io.open() is 100 times slower (!) on Python 2.6, and still about
     # three times slower on Python 2.7 (tested with "for _ in io.open(path): pass")
     return open(filename, mode) 
```
 If so, I'll make a commit with those changes.

marcelm: It’s not super important, but I’d suggest something like this (untested):
```python
if filename == '-':
     return _open_stdin_or_out(mode)
detected_format = _detect_format_from_extension(filename)
if detected_format is None and "w" not in mode:
    detected_format = _detect_format_from_content(filename)

if detected_format == "bz2":
    return _open_bz2(filename, mode)
elif detected_format == "xz":
    return _open_xz(filename, mode)
elif detected_format == "gz":
    return _open_gz(filename, mode, compresslevel, threads)
else:
    return open(filename, mode)
```
Then the first five lines deal with detecting the file format, and the rest deals with opening the file according to the detected format. It separates the concerns (detection vs opening) a bit better.

But this is essentially nitpicking. I just read the book *Clean Code*, which made me pay more attention to these types of thing :-).

src/xopen/__init__.py
```diff
@@ -356,6 +356,46 @@ def _open_gz(filename, mode, compresslevel, threads):
...
+            elif bs[:3] == b'\x42\x5a\x68':
+                # https://en.wikipedia.org/wiki/List_of_file_signatures
+                return _open_bz2(filename, mode)
+            elif bs[:6] == b'\xfd\x37\x7a\x58\x5a\x00' and _PY3:
+                # lzma module is not available for python 2.7
+                # https://tukaani.org/xz/xz-file-format.txt
+                return _open_xz(filename, mode)
+    except OSError as e:
+        # users expect IOError not OSError
+        raise IOError(e)
```
marcelm: As I wrote above, you should just return `None`, here but I wanted to mention anyway that `IOError` is an alias of `OSError` [starting from Python 3.3](https://docs.python.org/3/library/exceptions.html#OSError).

tests/file.txt.bz2.test
rhpvorderman: > To test whether the file format is recognized even though the file name extension is missing, we need (among others) a gzip file whose name does not end in .gz. 

Ah I see. Then adding links makes sense.

But since we can detect the file format by using bytes magic, does it make sense to keep the `.gz` recognition stuff? That is just extra maintenance load. It is been made redundant by the much more universal bytes magic.

src/xopen/__init__.py
bvaisvil: Thanks for the feedback. That looks good. I added a commit with those changes. Nitpicking is okay, it is your project after all!

tests/file.txt.bz2.test
marcelm: Yes, I think this could be changed to use content detection first if possible. However, we need `_detect_format_from_extension` anyway for when we’re writing to a file, and as a fallback for those cases where the file is actually not a regular file. For example, it could be a FIFO that has the `.gz` extension (I have done this, so it’s not entirely unrealistic ...). Let’s do that separately from this PR, though.

marcelm:
I’m happy with how this looks now, thanks a lot! We need to add something to the changelog, but I can do that later.

PR #32, opened 2020-06-18
Refacter xopen to allow for other formats to be opened with external programs.

rhpvorderman:
Since `gzip`, `pigz`, `bzip2` and `xz` all use a similar interface, the `PipedGzipReader` and `PipedGzipWriter` could very easily be generalized.

EDIT:
As a further addition the hardcoded compression level 6 has been removed, as well as the compression range check from 1 up til 9. This differs per tool.


src/xopen/__init__.py
```diff
@@ -85,23 +91,15 @@ def __del__(self):
             pass
 
 
-class PipedGzipWriter(Closing):
+class PipedCompressionWriter(Closing):
     """
-    Write gzip-compressed files by running an external gzip or pigz process and
```
marcelm, line 90: This docstring is now lost. It should be moved to the new `PipedGzipWriter`.

src/xopen/__init__.py
```diff
@@ -148,14 +148,9 @@ def _open_process(mode, compresslevel, threads, outfile, devnull):
         if sys.platform != 'win32':
             kwargs['close_fds'] = True
 
-        try:
```
marcelm, line 151: You’re switching from EAFP to LBYL here. That is, previously we would just try to start pigz and then gzip and then use the first one that worked, but now you’re first testing which program is available and then start it.

You may be able to convince me otherwise, but I’d prefer if the current way would be kept. It may not be so relevant in practice, but theoretically, between the time that `shutil.which` has determined which program is available and the time the program is actually started, the information on disk could have changed. Also, I think that `shutil.which` can give you false positives, that is, it could return the path to a binary that is not actually executable.

I think the way to solve this could be to add `PipedGzip...` and `PipedPigz...` classes, and then try them one after the other within `_open_gz`.

rhpvorderman: Fine with me! :+1:

rhpvorderman:
> I’ll be on vacation very soon, so in the worst case it may take a couple of weeks until I can get back to this.

No problem. I might not start on this immediately.

rhpvorderman:
Given the recent discovery of igzip I decided to put a bit of speed behind this.

I did not create a separate `PipedPigzWriter` class as that will significantly change behaviour compared to previous versions. I have simply rewritten the init to do try except.

rhpvorderman:
I have made a change necessary for igzip and other compression programs: the default compression is no longer hardcoded at level 6. Instead it uses `None` as default. In case of `None` no compression flag is passed to the process, effectively using its default level.

src/xopen/__init__.py
```diff
@@ -314,7 +333,10 @@ def _open_gz(filename, mode, compresslevel, threads):
     if 'r' in mode:
         return gzip.open(filename, mode)
     else:
-        return gzip.open(filename, mode, compresslevel=compresslevel)
+        # Override gzip.open's default of 9. That is way to slow for streaming
+        # to a large file.
```
marcelm, line 337:
```suggestion
        # Override gzip.open's default of 9 for consistency with command-line gzip
```

src/xopen/__init__.py
```diff
@@ -373,7 +395,8 @@ def xopen(filename, mode='r', compresslevel=6, threads=None):
     will raise an error.
 
     compresslevel is the compression level for writing to gzip files.
-    This parameter is ignored for the other compression formats.
+    This parameter is ignored for the other compression formats. If set to
+    None (default) will choose the default compression level for the format.
```
marcelm, line 399: That the default compression level is 6 for gzip should be mentioned somewhere because of the difference to the `gzip.open` default.

```suggestion
    None (default), level 6 is used.
```

src/xopen/__init__.py
```diff
@@ -85,23 +86,18 @@ def __del__(self):
...
-    par with gzip itself, but running an external gzip can still reduce wall-clock
-    time because the compression happens in a separate process.
+    Write Compressed files by running an external process and piping into it.
     """
-
-    def __init__(self, path, mode='wt', compresslevel=6, threads=None):
+    def __init__(self, path, program, mode='wt',
+                 compresslevel: Optional[int] = None,
+                 threads_flag: str = None,
+                 threads: Optional[int] = None):
```
marcelm, line 96: It’s ok to leave this now, but adding type annotation is somewhat unrelated and should have been a separate PR (also because it’s not done consistently in all changed places). Just leave it as it is now.

src/xopen/__init__.py
```diff
@@ -85,23 +86,18 @@ def __del__(self):
...
         """
         mode -- one of 'w', 'wt', 'wb', 'a', 'at', 'ab'
-        compresslevel -- gzip compression level
-        threads (int) -- number of pigz threads. If this is set to None, a reasonable default is
+        compresslevel -- compression level
+        threads (int) -- number of threads. If this is set to None, a reasonable default is
             used. At the moment, this means that the number of available CPU cores is used, capped
             at four to avoid creating too many threads. Use 0 to let pigz use all available cores.
```
marcelm, line 102: This still refers to pigz. Also, the `threads_flag` parameter is not documented.

src/xopen/__init__.py
```diff
@@ -114,11 +110,14 @@ def __init__(self, path, mode='wt', compresslevel=6, threads=None):
         self.devnull = open(os.devnull, mode)
         self.closed = False
         self.name = path
+        self.mode = mode
+        self.program = program
+        self.threads_flag = threads_flag
```
marcelm, line 115: Some or all of these attributes should probably be private.

src/xopen/__init__.py
```diff
@@ -295,9 +317,6 @@ def _open_bz2(filename, mode):
 
 
 def _open_xz(filename, mode):
-    if lzma is None:
```
marcelm, line 298: Removing this is unrelated and should have been in a separate PR (but it’s fine now).

marcelm (changes requested):
Hi, sorry for the delay – I was away longer than expected.

I just have a few small nits. Also, the title of the PR indicates otherwise, but this is just refactoring, right?

src/xopen/__init__.py
rhpvorderman: I removed the pigz references, moved the pigz specific stuff to the docstring for PipedGzipWriter. I also made the entire `PipedCompressionWriter` class private (`_PipedCompressionWriter`) and did the same for PipedCompressionReader.

rhpvorderman:
>  Also, the title of the PR indicates otherwise, but this is just refactoring, right? 

Yes. Initially I also added support for lzma and bzip2 formats, but your initial feedback suggested that it was better to do so in a separate PR. I have since updated the code, but had not updated the title. That is done now.

src/xopen/__init__.py
rhpvorderman: I put it back in. It should be there. It should also be there when a potential `PipedXzWriter` is used but for some reason is not available.

marcelm:
Thanks for the changes, but one more request: I think I’d prefer to not have the underscore in front of the `PipedCompressionReader`/`-Writer` classes since instances of that class are user-visible after all. It reminds me a bit of the [`_Record` class and others in PyVCF](https://pyvcf.readthedocs.io/en/latest/API.html#vcf-model-record), which one interacts with all the time. I believe they are only marked as private because one is not supposed to instantiate them directly. I think the situation here is comparable, and I’d like to not repeat that – IMO – mistake.

Also, I believe you added `src/xopen/resources/concatenated.gz.gz` by accident.

rhpvorderman:
In retrospect, without underscore indeed seems a lot better. Especially since it will be used as a type (in typed code) a lot. I made the change.

I also removed the resource. This was part of the igzip work and it accidentally slipped in.

marcelm (approved):

marcelm:
This looks good now, and it is a nice improvement!

PR #33, opened 2020-07-03
Igzip support

rhpvorderman:
Depends on merging #32. After that PR base branch can be switched to master.

~~Tests will fail because there is a bug in igzip https://github.com/intel/isa-l/issues/143/. This causes the append tests to fail because the gzipreader with igzip will produce the wrong output.~~

~~This PR will need to be on hold until the bug is fixed. Unfortunately I do not have the know-how to fix it myself. But I guess it will not take long for the upstream developers to fix.~~

EDIT: I see now that isal is not in ubuntu 18.04, but because of this issue we will p I robably have to use the conda version anyway. I have submitted the package to conda-forge.


rhpvorderman:
Hi Marcel,

I know you are on holiday, there is no rush. Enjoy the holiday!

 This morning I was thinking a bit about the bug in igzip that prevents reading of concatenated files. This is annoying, because I am not sure if my fastq files are one big gzip file or a concatenation of gzip files. So using igzip for reading is not safe. Even if the bug is fixed, we do not know if the user has installed the fixed version of igzip on their system.

Checking versions is very elaborate and annoying. But this morning I concluded that by simply testing a concatenated gzip file it would be much easier. So I have added a simple concatenated gzip file to the resources. If igzip can read it correctly, it is fixed. This way we can discriminate between correct and incorrect versions of igzip quite easily.  Writing concatenated gzip files with igzip is not bugged. So the PipedIGzipWriter can be used without issue.
 
By fixing it this way, we can immediately start using igzip. Writing zip files will become a lot faster. I have packaged the isa-l libraries in conda-forge, so if `isa-l` is added as a dependency of xopen in the next release we should start seeing the benefits. If the isa-l library is updated to include the fix for concatenated gzip files, this code will automatically become usable without any intervention, as well as being backwards compatible with older versions of igzip.

I have added the "DO NOT MERGE" label as this depends on #32 which should be merged first.

rhpvorderman:
Good things come to those who are patient: a fix has been made available to the igzip repo for the decompression bug. https://github.com/intel/isa-l/pull/147

This bug was found thanks to the extensive test suite of xopen, I am very grateful for that. 
Also igzip will be able to diminish compute time for compression level 1 of cutadapt by more than 40%, while delivering smaller filesize as well. Isn't that great? I am very grateful to the igzip program for that.

rhpvorderman:
@marcelm This is now ready for review.

src/xopen/__init__.py
```diff
@@ -67,6 +68,29 @@ def _available_cpu_count():
...
+    is much easier to use this, "proof in the pudding" type of evaluation.
+    The used test gzip file was created by:
+    echo -n "AB" | gzip -c > concatenated.gz
+    echo -n "CD" | gzip -c >> concatenated.gz
+    echo -n "EF" | gzip -c >> concatenated.gz
+    echo -n "GH" | gzip -c >> concatenated.gz
+    gzip and pigz can handle this without issue. But bugged version of igzip
+    will only extract the first "AB" block.
+    """
+    test_gz = pathlib.Path(__file__).parent / "resources" / "concatenated.gz"
```
marcelm, line 84: I would find it a bit nicer to create this file on the fly in a temporary directory instead of distributing it as a resource. Then you also don’t need to explain how you created the file in a comment since that will be documented right in the code. You also don’t need to concatenate four blocks, just two is enough.

Also, the remainder of the comment should be placed at the location where the function is called since it offers more general functionality than just checking igzip (even the name that you chose implies that already).

src/xopen/__init__.py
```diff
@@ -67,6 +68,29 @@ def _available_cpu_count():
...
+    try:
+        result = subprocess.run([program, "-c", "-d", str(test_gz.absolute())],
+                                check=True, stderr=PIPE, stdout=PIPE)
+    except (subprocess.CalledProcessError):
```
marcelm, line 88: I believe those parentheses aren’t necessary.

src/xopen/__init__.py
```diff
@@ -13,6 +13,7 @@
 import stat
 import signal
 import pathlib
+import subprocess
```
marcelm, line 16: Just noting, no change requested: Mixing `import subprocess` and `from subprocess import ...` is a bit inconsistent, but I can see why you wouldn’t want to import a function just named `run` at module level.

marcelm (commented):
Thanks, I think this looks good in general and will be a good addition.

I think what is missing is some kind of test of the different external tools. This doesn’t have to be in this PR because I think it would require some changes to how the test environment is setup, but it needs to be done since otherwise we will have more and more untested lines in the library.

src/xopen/__init__.py
rhpvorderman: Done.

rhpvorderman:
@marcelm. Thanks for the feedback. I have incorporated the changes. It is much better now with the temporary file creation.

I also added an igzip test by creating an igzip environment. In order to do so I switched the distro over to "focal" (20.04) which includes the `isal` package that includes igzip.

Sorry for my late reply, I have been busy with `python-isal` but more about that in a separate post. Also, for reasons that I do not want to highlight on the public internet, I will be offline for a while from some point in the coming weeks. I expect to be online again at the end of November. If I do not react to any of your posts it is because of this, and not out of rudeness, neglect or disinterest.

setup.py
```diff
@@ -20,6 +20,9 @@
     license='MIT',
     package_dir={'': 'src'},
     packages=find_packages('src'),
+    package_data={
```
marcelm, line 23: This can be removed.

src/xopen/__init__.py
```diff
@@ -93,6 +95,29 @@ def _set_pipe_size_to_max(fd: int):
     fcntl.fcntl(fd, fcntl.F_SETPIPE_SZ, _MAX_PIPE_SIZE)
 
 
+def _can_read_concatenated_gz(program: str) -> bool:
+    """
+    Check if a concatenated gzip file can be read properly. Not all deflate
+    programs handle this properly.
+    """
+    fd, temp_path = tempfile.mkstemp(suffix=".gz", prefix="xopen.")
```
marcelm, line 103: You could use `tempfile.NamedTemporaryFile` to simplify this a bit.

marcelm (changes requested):
Thanks again! I have two more comments, but if you don’t have time, I’ll make these changes myself and merge.

src/xopen/__init__.py
rhpvorderman: That would delete the file upon closing. And I want to close the file before I read it with igzip. Windows does not handle it well if two process open the same file (according to the Python documentation).

setup.py
rhpvorderman: :+1:

rhpvorderman:
@marcelm Thanks for the quick reply. It has been done.

src/xopen/__init__.py
marcelm: Ah yes, now I remember why I tend to use `TemporaryDirectory` most of the time ...

Issue #22, opened 2019-10-21
Making xopen more suitable for use with type annotation.


rhpvorderman:
Currently if I use 
```python
with open('my_file', 'rb') as file_handler:
    pass
``` 
I know that file_handler is of type `io.BufferedReader`. This allows me to easily find out which methods it has.

With xopen:

```python
with xopen('my_file', 'rb') as file_handler:
    pass
``` 

This can be any of several, including `io.BufferedReader`, `xopen.PipedGzipReader`, `lzma.LzmaFile` etc. These have some methods which are the same and some which are different.

It would be nice in the future to have four abstract base classes xopen. (Binary/Text)(Reader/Writer). So that when xopen is used you can type the resulting output file and know its methods.

This is not going to be trivial to implement but I would like to throw this idea out there. It will make coding easier for the 32 projects that are currently using it downstream. It is something on the wishlist after python 2 support is definitely dropped.

marcelm:
Do you mean to add type annotation suitable for type checkers and IDEs? I’d support that.

Note that the number of downstream projects using xopen is lower. Many of those listed by GitHub on the 'Used by' page are false positives, or use xopen indirectly through Cutadapt.

marcelm:
I think this was solved by f6c04d7

PR #40, opened 2020-11-08
Type annotations

marcelm:
This adds type annotations to almost all functions. The `xopen()` function itself is annotated as having the `IO` return type, which is not perfect, but should be good enough for now.

See #22

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,12 @@ jobs:
       script: flake8 src/ tests/
 
     - stage: test
+      name: mypy
+      python: "3.6"
+      install: python3 -m pip install mypy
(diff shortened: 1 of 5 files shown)


rhpvorderman (approved):
Looks good!

For consideration: removing 3.5 support (not officially supported now anymore anyway) simplifies the type annotation as variables can be annotated as well.
```myvar: str = "string``` instead of ```myvar = "string"  # type: str```

Also from 3.6 onwards there is the os.PathLike type.

marcelm:
Thanks for the review! And yes, let’s drop Python 3.5 very soon, also for f-strings!

PR #41, opened 2020-11-14
Fix ResourceWarnings when opening many piped processes

marcelm:
When many subprocesses are spawned (more than 50 in my tests), `fcntl.fcntl` starts to raise an OSError (permission denied). This is not handled in the PipedCompressionReader/-Writer __init__ methods, so the OSError is propagated and then the fallback mechanism finally opens the file using regular gzip.open. However, the Popen object is never closed correctly if this happens, leading to the ResourceWarning.

This fixes https://github.com/marcelm/cutadapt/issues/493

I’m opening this PR only for notification. I’ll need to merge this and make a new release soon.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -174,6 +174,7 @@ class PipedCompressionWriter(Closing):
 
         if threads is None:
             threads = min(_available_cpu_count(), 4)
+        self._threads = threads
         try:
             self.process = self._open_process(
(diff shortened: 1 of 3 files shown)


marcelm:
Travis passed, not sure why it doesn’t show up here.

rhpvorderman:
:+1: This was unforeseen. Good fix!

PR #42, opened 2020-12-08
Switch to GitHub actions

marcelm:
I recently got annoyed by Travis being so slow nowadays, so I switched to GitHub actions for Cutadapt. This is an adjusted version of that workflow.

This is even conceptually improved a bit because everything is now managed by tox. Also, this adds a test on macOS.

The deployment step uploads to the test PyPI server at the moment (I haven’t tested that, yet).


marcelm:
I changed it so the deploy job uploads to the real PyPI. We should test it by making a `dev` release before the actual release.

rhpvorderman (approved):
Looking good :+1:. It is nicer not to need a third-party service and Github has options to set up a private server as well when extra compute is needed (as soon as they decide charging money for running like travis did).

marcelm:
Agreed!

PR #43, opened 2020-12-16
Enable reading gzip files faster with the python main thread using python-isal.

rhpvorderman:
This adds python-isal to the mix as an optional dependency. Optional, because it is only available on linux and requires extra libraries to be present.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -28,6 +28,13 @@ except ImportError:
     lzma = None  # type: ignore
 
 try:
+    from isal import igzip
+    from isal import isal_zlib
+except ImportError:
+    igzip = None
+    isal_zlib = None
+
+try:
     import fcntl
     # fcntl.F_SETPIPE_SZ will be available in python 3.10.
     # https://github.com/python/cpython/pull/21921
@@ -490,11 +497,22 @@ def _open_gz(filename, mode: str, compresslevel, threads):
             pass  # We try without threads.
 
     if 'r' in mode:
+        if igzip is not None:
+            return igzip.open(filename, mode)
         return gzip.open(filename, mode)
-    else:
-        # Override gzip.open's default of 9 for consistency with command-line gzip.
-        return gzip.open(filename, mode,
-                         compresslevel=6 if compresslevel is None else compresslevel)
+
+    if igzip is not None:
+        try:
+            return igzip.open(filename, mode,
+                              compresslevel=isal_zlib.ISAL_DEFAULT_COMPRESSION
+                              if compresslevel is None else compresslevel)
+        except ValueError:
+            # Compression level not supported, move to built-in gzip.
+            pass
+
+    # Override gzip.open's default of 9 for consistency with command-line gzip.
+    return gzip.open(filename, mode,
+                     compresslevel=6 if compresslevel is None else compresslevel)
 
 
 def _detect_format_from_content(filename: str) -> Optional[str]:
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -477,22 +477,27 @@ def _open_xz(filename, mode: str) -> IO:
     return lzma.open(filename, mode)
 
 
+def _open_gz_external(filename, mode, threads, compresslevel):
+    if 'r' in mode:
+        try:
+            return PipedIGzipReader(filename, mode)
+        except (OSError, ValueError):
+            # No igzip installed or version does not support reading
+            # concatenated files.
+            return PipedGzipReader(filename, mode, threads=threads)
+    else:
+        try:
+            return PipedIGzipWriter(filename, mode, compresslevel)
+        except (OSError, ValueError):
+            # No igzip installed or compression level higher than 3
+            return PipedGzipWriter(filename, mode, compresslevel,
+                                   threads=threads)
+
+
 def _open_gz(filename, mode: str, compresslevel, threads):
     if threads != 0:
         try:
-            if 'r' in mode:
-                try:
-                    return PipedIGzipReader(filename, mode)
-                except (OSError, ValueError):
-                    # No igzip installed or version does not support reading
-                    # concatenated files.
-                    return PipedGzipReader(filename, mode, threads=threads)
-            else:
-                try:
-                    return PipedIGzipWriter(filename, mode, compresslevel)
-                except (OSError, ValueError):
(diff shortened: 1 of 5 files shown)


rhpvorderman:
Some benchmarks involving cutadapt in single-threaded mode:

Before:
```
$ hyperfine -w 3 -r  10 'cutadapt -a AGATCGGAAGAG 500000reads.fastq.qz -Z -o ramdisk/big2.cut.fastq.gz'
Benchmark #1: cutadapt -a AGATCGGAAGAG 500000reads.fastq.qz -Z -o ramdisk/big2.cut.fastq.gz
  Time (mean ± σ):      5.824 s ±  0.030 s    [User: 5.793 s, System: 0.029 s]
  Range (min … max):    5.785 s …  5.875 s    10 runs

$ du -h ramdisk/big2.cut.fastq.gz 
66M	ramdisk/big2.cut.fastq.gz

$ zcat ramdisk/big2.cut.fastq.gz | md5sum
97fc2a9c78e6597912a02b168e4cd068  -
```

After:
```
$ hyperfine -w 3 -r  10 'cutadapt -a AGATCGGAAGAG 500000reads.fastq.qz -Z -o ramdisk/big2.cut.fastq.gz'
Benchmark #1: cutadapt -a AGATCGGAAGAG 500000reads.fastq.qz -Z -o ramdisk/big2.cut.fastq.gz
  Time (mean ± σ):      4.010 s ±  0.067 s    [User: 3.973 s, System: 0.036 s]
  Range (min … max):    3.946 s …  4.190 s    10 runs

$ du -h ramdisk/big2.cut.fastq.gz 
62M	ramdisk/big2.cut.fastq.gz

$ zcat ramdisk/big2.cut.fastq.gz | md5sum
97fc2a9c78e6597912a02b168e4cd068  -
```
From 5.8 -> 4 seconds on average is quite a substantial improvement. Also the filesize is slightly smaller when using python-isal 66 -> 62m. So it is an approvement all around. Last but not least, the checksum of the content is the same, which is required.

This will be very very nice for people using cutadapt in single threaded mode. Galaxy users come to mind @bernt-matthias will probably like this.

If we get this into the conda release we can set the python-isal requirement for linux installs only. This will then be automatically available for biocontainers users as well on the next cutadapt release.

marcelm:
Wow, the results looks really good!

This should definitely go in, just let me find some time to review the code changes, which may take until after the holidays. I hope that’s fine.

tox.ini
```diff
@@ -12,20 +12,29 @@ commands =
...
 basepython = python3.6
 deps = flake8
 commands = flake8 src/ tests/
+skip_install = true
 
 [testenv:mypy]
 basepython = python3.6
 deps = mypy
-commands = mypy src/
+commands = mypy --ignore-missing-imports src/
```
marcelm, line 30: Using this command-line option is only recommended as [a last resort](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports). You should be able to just add a `# type: ignore` to the `from isal import ...` line instead.

tox.ini
```diff
@@ -12,20 +12,29 @@ commands =
...
+skip_install = true
 
 [flake8]
 max-line-length = 99
 max-complexity = 10
 extend_ignore = E731
+skip_install = true
```
marcelm, line 37: I think this line should not be added (this section is only read by `flake8`).

marcelm (changes requested):
Hi Ruben, I’ve had time for a review now and only have minor comments. 

I also was wondering about deployment. When this is merged, is it correct that the Conda package should be updated to depend on `python-isal` (on Linux)?

Also, for the `pip install` case, have you considered providing wheels for `python-isal`? Then one would no longer need to install `libisal-dev`.

Or you could go even one step further by linking statically against libisal (thus including it in the wheel), and then the user doesn’t need to use their distribution package manager at all.

tox.ini
rhpvorderman: Ah. Thanks! That is good to know. This missing import error is pretty annoying when using mypy and it is nice to know the proper fix.

tox.ini
rhpvorderman: Done

tox.ini
rhpvorderman: Done

src/xopen/__init__.py
```diff
@@ -27,6 +27,13 @@
 except ImportError:
     lzma = None  # type: ignore
 
+try:
+    from isal import igzip
+    from isal import isal_zlib
```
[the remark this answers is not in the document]
rhpvorderman: Done

rhpvorderman:
> I also was wondering about deployment. When this is merged, is it correct that the Conda package should be updated to depend on python-isal (on Linux)?

Yes, since that would propagate to the bioconda package for cutadapt as well. And as a result also the biocontainer. This will enable fast decompression/compression for a lot of users.

> Also, for the pip install case, have you considered providing wheels for python-isal? Then one would no longer need to install libisal-dev.

I have looked into this because of your comment. Initially I did not want to do it because it is a lot of hassle, and conda exists. I came to the conclusion that providing wheels adds no extra value. The isa-l libraries are still needed on the user system. So it does not eliminate a step for the user. I did however make the installation instructions more clear on the [python-isal readme](https://github.com/pycompression/python-isal) so it should be easier for users to install.

I changed the readme text on xopen to refer to this updated installation text.

> Or you could go even one step further by linking statically against libisal (thus including it in the wheel), and then the user doesn’t need to use their distribution package manager at all.

That would be a good solution. The dynamic linking works very well with conda and gives the benefit of automatically updating any bugs in isa-l away.
Static linking would however be more practical for more python users and would lead to less compatibility nightmares for me as the required version is simply included. The pace of change in isa-l is not that big and therefore it is very easy to keep up. There are however license issues that need to be discussed with the project maintainers before I am able to redistribute isa-l in that form. So I will do that.

marcelm:
I agree: If you provide wheels at all, they should include a statically linked isa-l.

Another reason: Many people still use `pip` to get Cutadapt (judging from the numbers for last month at https://pypistats.org/packages/cutadapt vs https://anaconda.org/bioconda/cutadapt/files), and almost all of these will not get the improvements automatically because `python-isal` is just an optional dependency. If there was a wheel, it could be made a hard dependency instead (using a suitable environment marker so you only require it on Linux).

PR #46, opened 2021-01-27
Make python-isal a requirement on macos as well.

rhpvorderman:
Wheels are now also build for macos as well on python-isal.
A twine check is run as part of the build for python-isal. Thanks for pointing me to cibuildwheel. It works great. 
https://github.com/pycompression/python-isal/blob/develop/.github/workflows/ci.yml
A lot of this can be copy pasted and applied to dnaio and cutadapt I think.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(
     package_data={"xopen": ["py.typed"]},
     extras_require={
         'dev': ['pytest'],
-        ':sys_platform=="linux" and python_implementation != "PyPy"': ['isal>=0.3.0']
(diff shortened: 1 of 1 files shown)


rhpvorderman:
Oh and by the way: thanks for including python-isal in xopen. Downloads on conda-forge have increased by 4x since that happened :smile: !
Also 6000 pypi downloads apparently :astonished: . Probably because cutadapt is used in CI environments though. Still, python-isal gets good testing that way :wink:.

marcelm:
Nice!

I’ll look into cibuildwheel again when I have some time. Would be good to get rid of the custom script.

PR #49, opened 2021-02-17
Use python -m isal.igzip as a replacement for pigz

### LINKED ISSUE #48: Replace pigz with python-isal where possible
Hi!

I was thinking about the weird dichotomy we have created. When you use multiple threads with xopen `pigz` will be used if `igzip` is not installed. Lately we have added python-isal support. This only works on one thread, but it is waaaaay more efficient than `pigz` in terms of compute time needed. On the other hand, this time is spent *inside* the python GIL. A disadvantage which pigz does not have. 
So when using multiple threads, pigz will be faster (if a low compression level is used) in wall clock time, but not in total cpu hours.

What to do about this? I didn't know. Then I realized that `python -m gzip` is a functional command. And that this behavior is also present in `python -m isal.igzip` (it's a complete drop-in replacement!). Unlike `python -m gzip` which uses a crazy buffer size of 1K (Idiocy! I should make a PR rectifying this). `python -m isal.igzip` uses `_compression.DEFAULT_BUFFER_SIZE` (8K on my system). It has piping capability. 

Check this out:

compression using pipes (similar to how xopen offloads processes)
```
$ hyperfine -w 2 -r 5 'cat big2.fastq | pigz -1 -p 1 > ramdisk/big2.fastq.gz'
Benchmark #1: cat big2.fastq | pigz -1 -p 1 > ramdisk/big2.fastq.gz
  Time (mean ± σ):     22.289 s ±  0.140 s    [User: 21.676 s, System: 0.849 s]
  Range (min … max):   22.174 s … 22.529 s    5 runs
 
$ hyperfine -w 2 -r 5 'cat big2.fastq | python -m isal.igzip -1 > ramdisk/big2.fastq.gz'
Benchmark #1: cat big2.fastq | python -m isal.igzip -1 > ramdisk/big2.fastq.gz
  Time (mean ± σ):      5.378 s ±  0.067 s    [User: 4.960 s, System: 0.992 s]
  Range (min … max):    5.294 s …  5.480 s    5 runs
 
$ hyperfine -w 2 -r 5 'cat big2.fastq | igzip -1 > ramdisk/big2.fastq.gz'
Benchmark #1: cat big2.fastq | igzip -1 > ramdisk/big2.fastq.gz
  Time (mean ± σ):      4.881 s ±  0.037 s    [User: 4.280 s, System: 0.775 s]
  Range (min … max):    4.861 s …  4.947 s    5 runs
```

Decompression using pipes
```
$ hyperfine -w 2 -r 5 'cat big2.fastq.gz | pigz -d -p 1 > ramdisk/big2.fastq'
Benchmark #1: cat big2.fastq.gz | pigz -d -p 1 > ramdisk/big2.fastq
  Time (mean ± σ):      6.644 s ±  0.020 s    [User: 6.269 s, System: 0.584 s]
  Range (min … max):    6.629 s …  6.680 s    5 runs
 
$ hyperfine -w 2 -r 5 'cat big2.fastq.gz | python -m isal.igzip -d  > ramdisk/big2.fastq'
Benchmark #1: cat big2.fastq.gz | python -m isal.igzip -d  > ramdisk/big2.fastq
  Time (mean ± σ):      4.473 s ±  0.022 s    [User: 3.912 s, System: 0.843 s]
  Range (min … max):    4.454 s …  4.506 s    5 runs
 
$ hyperfine -w 2 -r 5 'cat big2.fastq.gz | igzip -d  > ramdisk/big2.fastq'
Benchmark #1: cat big2.fastq.gz | igzip -d  > ramdisk/big2.fastq
  Time (mean ± σ):      3.414 s ±  0.030 s    [User: 2.835 s, System: 0.634 s]
  Range (min … max):    3.377 s …  3.451 s    5 runs
```

So if `igzip` is NOT installed (very likely) using  `python -m isal.igzip` is the fastest way on systems where python-isal is installed (100% guaranteed on linux and mac).

Are you thinking what I am thinking? `git checkout -b MOARRRPERFORMANCE1!1!` :rocket:

rhpvorderman:
fixes #48 

Here are some benchmarks with a real-world scenario: cutadapt.

Using igzip (only available via conda installation):
```
$ hyperfine -w 3 -r  10 'cutadapt -a AGATCGGAAGAG 5000000reads.fastq.gz -Z --cores 4 -o ramdisk/big2.cut.fastq.gz'
Benchmark #1: cutadapt -a AGATCGGAAGAG 5000000reads.fastq.gz -Z --cores 4 -o ramdisk/big2.cut.fastq.gz
  Time (mean ± σ):      8.731 s ±  0.083 s    [User: 42.513 s, System: 3.043 s]
  Range (min … max):    8.608 s …  8.864 s    10 runs
```
Wall clock time: 8.7 sec. Compute time: 45.5 sec.

Using pigz. Generally available (gzip availability is even better, but performs worse):
```
$ hyperfine -w 3 -r  10 'cutadapt -a AGATCGGAAGAG 5000000reads.fastq.gz -Z --cores 4 -o ramdisk/big2.cut.fastq.gz'
Benchmark #1: cutadapt -a AGATCGGAAGAG 5000000reads.fastq.gz -Z --cores 4 -o ramdisk/big2.cut.fastq.gz
  Time (mean ± σ):     10.245 s ±  0.071 s    [User: 77.794 s, System: 3.061 s]
  Range (min … max):   10.156 s … 10.382 s    10 runs
```
Wall clock time: 10.2 sec. Compute time: 80.8 sec.
The wall clock increase is not too bad. But do note this is a 6 core 12 thread system. On a 4-thread system the wall-clock time would have scaled linearly with the compute time.

Using `python -m isal.igzip`. Due to the latest changes available on Linux and Mac if xopen is installed (== always from our perspective, unless PyPy is used.).
```
$ hyperfine -w 3 -r  10 'cutadapt -a AGATCGGAAGAG 5000000reads.fastq.gz -Z --cores 4 -o ramdisk/big2.cut.fastq.gz'
Benchmark #1: cutadapt -a AGATCGGAAGAG 5000000reads.fastq.gz -Z --cores 4 -o ramdisk/big2.cut.fastq.gz
  Time (mean ± σ):      9.585 s ±  0.050 s    [User: 47.314 s, System: 4.087 s]
  Range (min … max):    9.506 s …  9.642 s    10 runs
```
Wall clock time: 9.6 sec. Compute time 51.4 sec.
There is a slight hit in wall clock and compute time compared to igzip, but much better than pigz.

This was tested with the following branch of python-isal: https://github.com/pycompression/python-isal/pull/38 . That will need to be released before this change can be used (adds a '-c' flag).

If this gets released cutadapt users who install via PyPI and use multiple cores will see a compute time decrease of ~35%. Conda and biocontainer users of cutadapt will see no additional benefit.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -9,7 +9,6 @@ import sys
 import io
 import os
 import bz2
-import time
 import stat
 import signal
 import pathlib
@@ -298,9 +297,9 @@ class PipedCompressionReader(Closing):
         assert self.process.stderr is not None
         self._stderr = io.TextIOWrapper(self.process.stderr)
         self.closed = False
-        # Give the subprocess a little bit of time to report any errors (such as
-        # a non-existing file)
-        time.sleep(0.05)
+        # Peek the first character(s) of stdout. This will ensure the program
+        # has started before checking any errors.
+        self.process.stdout.peek(1)  # type: ignore  # stdout is io.BufferedReader if set to PIPE.
         self._raise_if_error()
 
     def __repr__(self):
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -17,7 +17,7 @@ import subprocess
 import tempfile
 from abc import ABC, abstractmethod
 from subprocess import Popen, PIPE, DEVNULL
-from typing import Optional, TextIO, AnyStr, IO
+from typing import Optional, TextIO, AnyStr, IO, List
 
 from ._version import version as __version__
 
@@ -155,9 +155,9 @@ class PipedCompressionWriter(Closing):
     """
     Write Compressed files by running an external process and piping into it.
     """
-    def __init__(self, path, program: str, mode='wt',
+    def __init__(self, path, program_args: List[str], mode='wt',
                  compresslevel: Optional[int] = None,
-                 threads_flag: str = None,
+                 threads_flag: Optional[str] = None,
                  threads: Optional[int] = None):
         """
         mode -- one of 'w', 'wt', 'wb', 'a', 'at', 'ab'
@@ -174,11 +174,11 @@ class PipedCompressionWriter(Closing):
 
         # TODO use a context manager
         self.outfile = open(path, mode)
-        self.closed = False
-        self.name = path
-        self._mode = mode
-        self._program = program
-        self._threads_flag = threads_flag
+        self.closed: bool = False
+        self.name: str = path
+        self._mode: str = mode
+        self._program_args: List[str] = program_args
+        self._threads_flag: Optional[str] = threads_flag
 
         if threads is None:
             threads = min(_available_cpu_count(), 4)
@@ -202,14 +202,14 @@ class PipedCompressionWriter(Closing):
             self.__class__.__name__,
             self.name,
             self._mode,
-            self._program,
+            " ".join(self._program_args),
             self._threads,
         )
 
     def _open_process(
         self, mode: str, compresslevel: Optional[int], threads: int, outfile: TextIO,
     ) -> Popen:
-        program_args = [self._program]
+        program_args: List[str] = self._program_args[:]  # prevent list aliasing
         if threads != 0 and self._threads_flag is not None:
             program_args += [self._threads_flag, str(threads)]
         extra_args = []
@@ -240,7 +240,8 @@ class PipedCompressionWriter(Closing):
         self.outfile.close()
         if retcode != 0:
             raise OSError(
-                "Output {} process terminated with exit code {}".format(self._program, retcode))
+                "Output {} process terminated with exit code {}".format(
+                    " ".join(self._program_args), retcode))
 
     def __iter__(self):  # type: ignore
         # For compatibility with Pandas, which checks for an __iter__ method
@@ -259,7 +260,7 @@ class PipedCompressionReader(Closing):
     def __init__(
         self,
         path,
-        program: str,
+        program_args: List[str],
         mode: str = "r",
         threads_flag: Optional[str] = None,
         threads: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


src/xopen/__init__.py
```diff
@@ -383,9 +384,9 @@ class PipedGzipReader(PipedCompressionReader):
     """
     def __init__(self, path, mode: str = "r", threads: Optional[int] = None):
         try:
-            super().__init__(path, "pigz", mode, "-p", threads)
+            super().__init__(path, ["pigz"], mode, "-p", threads)
```
marcelm, line 387: We should probably (in a different PR) split this up into a `PipedGzipReader` and a `PipedPigzReader` (same for the `...Writer`) because that would be more consistent with the other classes: They run the tool or fail, and any fallback to a different tool is handled outside the class.

marcelm:
I think this looks good and should go in (when the tests pass). Referring to my comment in issue #48, I can then later figure out how to use this in Cutadapt. It’s an improvement no matter what.

Regarding one of the test failures: gzip files don’t necessarily have a `.gz` extension. For example, `.bam` files are gzip-compressed, and some people use `.bgz` for bgzip-compressed files (which can be decompressed by regular gzip). So I wouldn’t require the extension.

rhpvorderman:
Yes, this is an artifact of python-isal striving to be a reimplementation of the gzip module. Unfortunately `python -m gzip` is deeply flawed. As I keep on discovering. I can fix this trivially in a new release. I will come back to this.

rhpvorderman:
The tests are now fixed. Using `time.sleep` to wait a bit for a possible error turned out to be a liability. Sometimes an error was not produced in time. The wait of `0.01` second was basically an eternity when starting up `pigz`, `gzip` or `igzip` but `python -m isal.igzip` starts an entire python interpreter. There is much more latency there. This latency will vary wildly on systems (hard disk, may be a VM) and caused issues on the github actions VMs even when the timeout was increased.

`time.sleep` has now been replaced with `self.process.stdout.peek(1)` to force reading of the output. Which forces the program to run (and crash). It is not sensitive to the speed of the machine.

rhpvorderman:
Another advantage of this PR is that python-isal is theoretically portable to Windows (haven't got the time yet to do it). This will enable the enhanced compression speed once it is done in all modes of xopen for windows users to. Igzip does not run on Windows, and since it relies on unix system calls it will not run on windows unless someone gives it some love. Python-isal does not have this problem, but some effort is required to make it build.

marcelm:
I like the `.peek()` solution, very nice.

PR #51, opened 2021-02-23
Split PipedGzipReader/-Writer classes

### LINKED ISSUE #50: Split PipedGzipReader/-Writer
We should factor out a `PipedPigzReader` out of `PipedGzipReader` to be more consistent with the other classes: They either run the tool or fail. Fallback to a different tool needs to be handled outside the class.

Same for `PipedGzipWriter`.

marcelm:
This closes #50 by splitting a PipedPigzReader/-Writer class off from PipedGzipReader/-Writer.

I also refactored the tests a bit to ensure that all the Piped... classes are tested, not only PipedGzipReader.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -590,16 +590,17 @@ def xopen(
     bzip2 and xz. If the filename is '-', standard output (mode 'w') or
     standard input (mode 'r') is returned.
 
-    The file type is determined based on the filename: .gz is gzip, .bz2 is bzip2, .xz is
-    xz/lzma and no compression assumed otherwise.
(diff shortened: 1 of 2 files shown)


rhpvorderman:
Hi, sorry, seems like I worked on the same thing at the same time.
I think #52 can be merged into this one. The changes to the writers and readers are basically the same, but the `_open_external_gz` function is slightly improved (IMHO) in my PR, while the testing is not touched there.

rhpvorderman:
Nevermind, that change caused too many headaches #52 is closed.

rhpvorderman (approved):
Looks great!

PR #52, opened 2021-02-23
Put every reader and writer in its own class, better _gz_open_external function

rhpvorderman:
Also the reader and writer classes are  now organized below eachother


rhpvorderman:
Nevermind. This change causes errors and headaches.

marcelm:
Yeah, I also had the desire of simplifying the `_open_external_gz` function, but the cases were just too different.

PR #53, opened 2021-02-23
Obtain coverage for full source, not only tests

marcelm:
Since commit 8b71eb30d31b2c1e98c22409f8ab5d733988e5c2, coverage reports only included the `test_xopen.py` module, not the source of the `xopen` module.

The test fails because we had artificially inflated coverage until now.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@ deps =
     coverage
 setenv = PYTHONDEVMODE = 1
 commands =
-    coverage run --concurrency=multiprocessing -m pytest --doctest-modules src/xopen/ tests/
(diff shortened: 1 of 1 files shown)


rhpvorderman:
I have a codecov.yml file in the python-isal repository: https://github.com/pycompression/python-isal/blob/develop/codecov.yml
A drop in coverage is acceptable as long as the patch is properly covered and it does not lower the project coverage below a certain level. We could add that as well at some point.

rhpvorderman:
Looks good btw!

marcelm:
Yes, I have such a file for Cutadapt as well, but I was happy with the defaults for xopen until now and didn’t want to clutter the repository with another file. I think it’s appropriate that the test failed for this PR because the drop in coverage was so big.

PR #54, opened 2021-02-24
Add extra tests to increase coverage

rhpvorderman:
Coverage was inflated still because the files in tests were included. These files do not matter for coverage. Coverage is now run with `--source=xopen` so only the actual package is tested.

I added a html report to the coverage reporting. It makes it easy to check coverage. I added some tests for uncovered lines.
Consequently I found a bug in the `peek` method, which can only read binary files. I also removed some legacy code to increase the coverage (fspath workaround).


marcelm:
> Coverage was inflated still because the files in tests were included.

Test files absolutely need to be included! Tests are code as well and the same mistakes can happen as in "normal" code. For example, I have written tests that were never executed because I accidentally had two functions with the same name.

tests/test_xopen.py
```diff
@@ -21,7 +23,7 @@
     PipedPythonIsalWriter,
     _MAX_PIPE_SIZE,
     _can_read_concatenated_gz,
-    igzip,
+    igzip
```
marcelm, line 26: Please leave in the trailing comma. Trailing commas reduce diff size. We don’t use [Black](https://black.readthedocs.io/en/stable/the_black_code_style.html), yet, but it would add the trailing comma.

tests/test_xopen.py
```diff
@@ -504,3 +514,80 @@ def test_open_many_gzip_writers(tmp_path):
...
+        PipedCompressionWriter("test", ["gzip"], "xb")
+    error.match("Mode is 'xb', but it must be")
+
+
+def test_pipedcompressionwriter_wrong_program():
+    with pytest.raises(OSError):
+        PipedCompressionWriter("test", ["XVXCLSKDLA"], "wb")
+
+
+def test_compression_level(tmpdir, gzip_writer):
```
marcelm, line 530: We should switch to using `tmp_path` at some point. (Not in this PR.)

marcelm (changes requested):
Cool, this looks good. Removing some of the Python 3.5 stuff was really necessary!

rhpvorderman:
>  For example, I have written tests that were never executed because I accidentally had two functions with the same name.

Wouldn't that show up as a reduced line coverage in the actual code?
Anyway, I enabled the tests code for coverage again.

tests/test_xopen.py
rhpvorderman: :+1: Pathlib.Path works great when writing tests.

tests/test_xopen.py
rhpvorderman: Done.

tox.ini
```diff
@@ -7,10 +7,11 @@ deps =
     coverage
 setenv = PYTHONDEVMODE = 1
 commands =
-    coverage run --concurrency=multiprocessing -m pytest --doctest-modules --pyargs xopen tests
```
marcelm, line 10: I also just noticed that we have coverage configuration in both `setup.cfg` and `tox.ini` (my fault). We can get rid of `setup.cfg` actually because the only other thing in it is the "universal wheel" configuration, and we don’t need that since we’re not compatible with Python 2 anymore.

tox.ini
```diff
@@ -7,10 +7,11 @@ deps =
...
+    coverage run -m pytest --doctest-modules tests
     coverage combine
```
marcelm, line 11: `coverage combine` is only necessary when running with `--concurrency=multiprocessing` or `-p`, so this can be removed as well (if `parall = True` is also removed from the configuration).

marcelm:
> Wouldn't that show up as a reduced line coverage in the actual code?

Perhaps, but line coverage isn’t foolproof, so it could just reduce branch coverage, or not even that.

tox.ini
rhpvorderman: Ah, I was wondering why my usual coverage fu was not working. 
I removed setup.cfg and updated tox.ini accordingly.

tox.ini
rhpvorderman: I removed setup.cfg.

marcelm:
Perhaps you could enable branch coverage while you’re at it? Should be as simple as adding `branch = True` under the `[coverage:run]` section in `tox.ini`.

rhpvorderman:
@marcelm, oooh those graphs look nice. Strange that this is not enabled by default.

marcelm:
Totally agree!

marcelm (approved):

marcelm:
And thanks for cleaning up the coverage configuration. I don’t remember why it ended up looking the way it did, but it’s much simpler now.

marcelm:
Merge ahead when you’re done, I don’t want to do it now in case you’re still changing something, but I need to work on something else.

rhpvorderman:
No problem. These things just happen on old projects. Sometimes some good old refactoring is necessary. And I enjoy removing lines from a project. The less code the better, less maintenance! :wink:

PR #55, opened 2021-03-10
Install isal wheels on windows too.

rhpvorderman:
I enabled installation on windows for python-isal. Wheels are provided as well: https://pypi.org/project/isal/0.6.0/#files

@marcelm You may want to take a loot at the ci.yml with cibuildwheel of python-isal. I think it will be quite trivial for you to create windows and mac wheels as well for dnaio and cutadapt.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ setup(
     package_data={"xopen": ["py.typed"]},
     extras_require={
         'dev': ['pytest'],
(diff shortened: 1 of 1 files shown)


rhpvorderman:
Whoops. I fixed a bug in CPython `python -m gzip`. Ported the bugfix back to `python -m isal.igzip`. But that fix makes xopen crash. patching python-isal now.

PR #57, opened 2021-03-30
Always install isa-l on supported 64-bit systems.

rhpvorderman:
Pypy is now supported. Wheels are also build for aarch64.

isa-l is 64-bit only.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,12 @@ setup(
     package_data={"xopen": ["py.typed"]},
     extras_require={
         'dev': ['pytest'],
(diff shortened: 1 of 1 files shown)


rhpvorderman:
Unfortunately due to wheels being built on all relevant 64-bit platforms, that makes it harder to test the fallback code. Which isn't covered anymore.
We could set some environment variable to disable isal installation maybe?

marcelm:
Let’s merge this and improve the test coverage separately, see #58.

PR #59, opened 2021-04-02
Add tests for no isal installation

### LINKED ISSUE #58: Fallback code when isal unavailable is not tested
#57 reduced the test coverage, and we need to fix this.

The `isal` module is now always imported succesfully in all test configurations.

My suggestion is to
1. change `if igzip is not None` to `if "isal" not in sys.modules`,
2. and use [pytest.monkeypatch](https://docs.pytest.org/en/stable/monkeypatch.html) to temporarily remove the `isal` module from `sys.modules` during a test, forcing use of the fallback code.

rhpvorderman:
fixes #58 

This restores the coverage after #57 . Except for the lines after `except ImportError` when isal is imported, but these are so simple (set two variables to None) that testing is not really needed.

EDIT: https://app.codecov.io/gh/pycompression/xopen/compare/59/changes#D1L42 These lines can probably not be recovered for coverage. Unless we start testing on 32-bit systems that will never support isal. Or architectures other than x86_64 and aarch64.

### DIFF
--- a/tests/test_xopen.py
+++ b/tests/test_xopen.py
@@ -557,7 +557,7 @@ def test_xopen_fals_back_to_gzip_open_write_no_isal(lacking_pigz_permissions,
     tmp = tmp_path / "test.gz"
     with xopen.xopen(tmp, "wb") as f:
         f.write(b"hello")
-    assert gzip.decompress(tmp.read_bytes())== b"hello"
+    assert gzip.decompress(tmp.read_bytes()) == b"hello"
 
 
 def test_open_many_gzip_writers(tmp_path):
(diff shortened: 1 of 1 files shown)


tests/test_xopen.py
```diff
@@ -145,6 +145,24 @@ def test_xopen_binary(fname):
         assert lines[1] == b'The second line.\n', fname
 
 
+def test_xopen_binary_no_isal_no_threads(fname, monkeypatch):
+    import xopen  # xopen local overrides xopen global variable
+    monkeypatch.setattr(xopen, "igzip", None)
```
marcelm, line 150: Can you create a fixture (`xopen_without_igzip` or so) that does the `import xopen` and `monkeypatch.setattr(...)`? That would eliminate some of the code duplication.

rhpvorderman: Good suggestion. It is done.

marcelm:
Nice!

PR #61, opened 2021-04-07
Add windows test to CI

rhpvorderman:
Since we are busy increasing coverage...


rhpvorderman:
Seems like I opened a huge can of worms. This should be revisited later.

rhpvorderman:
@marcelm Are you testing this via the CI? (Given the quite large amount of WIP comments, that seems likely). 
I find that MacOS and Linux are similar enough that this is usually successful, but for windows it is much quicker to boot up windows and test it there.

I do have a windows installation on my machine. I even have development environments set up there. (For python-isal I had to do quite some tedious work to get it to install, but in the end I got it to work). 
Now I'd rather play games on that Windows install, but I don't mind debugging an hour in my spare time to get xopen to work on windows :slightly_smiling_face: .

src/xopen/__init__.py
```diff
@@ -358,6 +366,11 @@ def _raise_if_error(self, check_allowed_code_and_message: bool = False,
         """
         retcode = self.process.poll()
 
+        if sys.platform.startswith("win") and retcode == 1 and stderr_message == b'':
+            # Special case for Windows. Winapi terminates processes with exit code 1
+            # and an empty error message.
+            return
```
rhpvorderman, line 372: @marcelm Can you have a look at this? 
The following is the case.
The winapi terminates processes with exit code one by default, and not with signal.SIGTERM. (This is a bug IMO, but what can you do about it?). 

If python -m isal.igzip crashes on windows it will also give exit code 1. (Only external application readily available on windows). Luckily that will *always* fail verbosely, due to the nature of the python interpreter. By checking for platform,  exit code and empty error message I think we can be quite sure it works correctly.

However this has to be implemented outside the "_allowed_exit_code" and "_allowed_error_message" checks in order to function correctly.

rhpvorderman:
@marcelm I did it! All tests succeed. Hurray!

marcelm:
Hi, thanks for picking this up!

> Are you testing this via the CI?

Yes, I was, and this is indeed where the "wip" commits come from, but I switched to my Windows VM for the last couple of commits.

I’ll force-push a somewhat cleaner commit history.

src/xopen/__init__.py
marcelm: Ah, that was the last remaining thing that I didn’t have time to figure out, thanks for that. (Probably I wouldn’t have gone through the trouble of installing PyCharm in my Windows VM, so it would have taken me much longer.)

Yes, let’s handle it the way you did it.

I’ll change `sys.platform.startswith("win")` to `sys.platform == "win32"`, however, because that’s how we wrote it in other parts of the code. This is fine because `sys.platform` is guaranteed to be "win32" even on a 64-bit Windows.

rhpvorderman:
It looks good to me. Unless you feel something should be added, it can be merged.

Issue #16, opened 2019-08-14
Investigate test_truncated_gz failure


marcelm:
Possibly the simplification of `PipedGzipReader.read` introduced in #14 sometimes causes a test failure, as [seen on Travis](https://travis-ci.org/marcelm/xopen/builds/571788800).

rhpvorderman:
Hmmm. I can run this through hyperfine with a 100 runs. Using conda I can test multiple interpreters.
If that helps.

marcelm:
I can now actually reproduce this locally with `while tox; do true; done`. This failed after a minute or so.

marcelm:
I think it’s a timing issue in `PipedGzipReader.close()`.

rhpvorderman:
I cannot reproduce the error on my machine. Which makes me suspect that it is a race condition, and this machine is just to fast to get it.
I made a small fix in #17 which should prevent the race condition from happening.

marcelm:
I also thought that adding a `.wait()` would be an idea, but there’s a [big warning in the docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) that it can deadlock when used with `stdout=PIPE`, which we do use. The intention of the previous code within `read()` and `readline()` was to trigger the `.wait()` only when we were sure that we were already at EOF, which would imply that the process won’t write any further data into the kernel buffer.

I’m not sure that a `self.process.terminate()` followed by an immediate `self.process.wait()` cannot deadlock. When you read just a couple of bytes from a file and then close it before reaching the end, perhaps the signal gets queued and delivered only after the process has managed to fill the buffer already.

However, I don’t see this happening in my tests and the PR fixes this particular issue.

rhpvorderman:
> I also thought that adding a `.wait()` would be an idea, but there’s a [big warning in the docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) that it can deadlock when used with `stdout=PIPE`, which we do use. The intention of the previous code within `read()` and `readline()` was to trigger the `.wait()` only when we were sure that we were already at EOF, which would imply that the process won’t write any further data into the kernel buffer.

When working on [pytest-workflow](https://github.com/lumc/pytest-workflow) I have experienced issues with `.wait()` because I was piping stdout and sterr of a workflow engine to a PIPE and the process hanged in production because long pipelines produce more stdout than 64kb. 

Apperently Popen.communicate() can fix this. I used it successfullty with pytest-workflow. 
[...]

PR #63, opened 2021-04-30
Enable the use of pbzip2.

DriesSchaumont:


src/xopen/__init__.py
```diff
@@ -202,7 +206,10 @@ def _open_process(
     ) -> Popen:
         program_args: List[str] = self._program_args[:]  # prevent list aliasing
         if threads != 0 and self._threads_flag is not None:
-            program_args += [self._threads_flag, str(threads)]
+            if self._space:
+                program_args += [self._threads_flag, str(threads)]
+            else:
+                program_args += [f"{self._threads_flag}{threads}"]
```
rhpvorderman, line 212: This if else clause is making the program more complex than it should be. Currently the only other program accepting threads is `pigz` and it also parses `-p4` correctly.

So it is best IMO to remove all the `_space` stuff and simply change the line on 205 with the line on 212. No if clauses needed.

src/xopen/__init__.py
```diff
@@ -273,7 +281,10 @@ def __init__(
                 #   using multiple threads while there is only a 10% gain in wall
                 #   clock time.
                 threads = 1
-            program_args += [threads_flag, str(threads)]
+            if space:
+                program_args += [threads_flag, str(threads)]
+            else:
+                program_args += [f"{threads_flag}{threads}"]
```
rhpvorderman, line 287: Same as above.

src/xopen/__init__.py
```diff
@@ -441,6 +452,33 @@ def __init__(
...
+    def __init__(
+        self,
+        path,
+        mode: str = "wt",
+        compresslevel: Optional[int] = None,
+        threads: Optional[int] = None,
+    ):
+        """
+        """
+        if compresslevel is not None and compresslevel not in self._accepted_compression_levels:
```
rhpvorderman, line 478: can also be `compresslevel not in range(10)`. Much simpler. Pigz however accepts (0-9, 11) so it needed more complex code.

rhpvorderman (changes requested):
Looks good! Thanks for the PR. I think the code can be simplified somewhat.

src/xopen/__init__.py
rhpvorderman: For clarification, just keep the old code the same, but replace `program_args += [self._threads_flag, str(threads)]` with `program_args += [f"{self._threads_flag}{threads}"]`. That solves the space problem and will not cause problems for pigz. Also this makes the `space` parameter that was added redundant. Doing it this way keeps it simpler.

rhpvorderman:
I will be on holiday for the coming two weeks, but I am quite sure that won't hamper this review process.

src/xopen/__init__.py
DriesSchaumont: Ah, of course! Adjusted.

src/xopen/__init__.py
DriesSchaumont: Done!

src/xopen/__init__.py
```diff
@@ -441,6 +443,32 @@ def __init__(
...
+
+    def __init__(
+        self,
+        path,
+        mode: str = "wt",
+        compresslevel: Optional[int] = None,
+        threads: Optional[int] = None,
+    ):
+        """
+        """
```
marcelm, line 467: Please remove the empty docstring (or write something)

src/xopen/__init__.py
```diff
@@ -499,7 +527,16 @@ def _open_stdin_or_out(mode: str) -> IO:
     return open(std.fileno(), mode=mode, closefd=False)
 
 
-def _open_bz2(filename, mode: str) -> IO:
+def _open_bz2(filename, mode: str, compresslevel, threads):
```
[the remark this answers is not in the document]
marcelm: I’m not sure that `compresslevel` should actually be handled, at least not in this way. When opening a bz2 file through `bz2.open`, we ignore the `compresslevel` at the moment.

The compression level for Bzip2 files is on quite a different scale than for gzip files. bzip2 and pbzip2 compress at level 9 (block size 900k) by default, and reducing it to 1 (block size 100k) changes runtime and compressed size only a bit (14% slower, 7% bigger in a short test I did, difference in runtime is a bit bigger than that for `pbzip2`).

Since `xopen()` can handle multiple file formats, it may be necessary that it gets another parameter for just the bzip2 compression level if we want full flexibility. But the easiest thing to do for the moment IMO would be to just behave the same as before and just *not* set the compression level.

marcelm (commented):
Thanks for this! This has been on my nice-to-have list for a while.

I noticed that the tests did not run automatically when you submitted the PR. See
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks (must be because of the bot PRs abusing GitHub infrastructure for cryptocurrency mining). I will start the tests now, and there will be a couple of style issues that should also be fixed.

src/xopen/__init__.py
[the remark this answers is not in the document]
DriesSchaumont: I removed the parsing of compresslevel and added the type hint!

src/xopen/__init__.py
DriesSchaumont: Done!

src/xopen/__init__.py
```diff
@@ -254,7 +256,7 @@ def __init__(
         program_args: List[str],
         mode: str = "r",
         threads_flag: Optional[str] = None,
-        threads: Optional[int] = None,
+        threads: Optional[int] = None
```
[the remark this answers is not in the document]
DriesSchaumont: Adjusted

DriesSchaumont:
I added some tests. Please let me know if this increased the diff size a bit too much.

tests/test_xopen.py
```diff
@@ -223,20 +307,25 @@ def test_readline_text(fname):
         assert f.readline() == CONTENT_LINES[0]
 
 
-def test_gzip_reader_readline(gzip_reader):
+def test_reader_readline(reader):
```
marcelm, line 310: I like this approach much better: You changed a function that just tested gzip files to now also test bzip2 files. Can you go through the other places where you duplicated code and see whether you can use them same idea?

tests/test_xopen.py
```diff
@@ -392,6 +492,14 @@ def test_truncated_gz(truncated_gzip):
             f.close()  # pragma: no cover
 
 
+def test_truncated_bz2(truncated_bz2):
```
marcelm, line 495: To eliminate the duplication, it would be nice to switch to a `truncated_compressed_file` fixture or something like that.

tests/test_xopen.py
```diff
@@ -442,17 +574,17 @@ def test_write_pigz_threads_no_isal(tmpdir, xopen_without_igzip):
...
-def test_read_gzip_no_threads():
-    import gzip
-    with xopen("tests/hello.gz", "rb", threads=0) as f:
-        assert isinstance(f, gzip.GzipFile), f
-
-
-def test_write_gzip_no_threads(tmpdir):
-    import gzip
-    path = str(tmpdir.join("out.gz"))
+def test_write_no_threads(tmpdir, ext):
```
marcelm, line 577: This is much nicer than before!

tests/test_xopen.py
```diff
@@ -442,17 +574,17 @@ def test_write_pigz_threads_no_isal(tmpdir, xopen_without_igzip):
...
+    klasses = {
+        ".bz2": bz2.BZ2File,
+        ".gz": gzip.GzipFile,
+        ".xz": lzma.LZMAFile,
+        "": io.BufferedWriter
```
marcelm, line 582: Please add a trailing comma

src/xopen/__init__.py
```diff
@@ -662,6 +697,6 @@ def xopen(
     elif detected_format == "xz":
         return _open_xz(filename, mode)
     elif detected_format == "bz2":
-        return _open_bz2(filename, mode)
+        return _open_bz2(filename, mode, compresslevel, threads)
```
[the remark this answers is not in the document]
marcelm: ... and you fixed it while I wrote this.

tests/test_xopen.py
```diff
@@ -69,6 +73,30 @@ def available_gzip_readers_and_writers():
 PIPED_GZIP_READERS, PIPED_GZIP_WRITERS = available_gzip_readers_and_writers()
 
 
+def available_bzip2_readers_and_writers():
+    readers = [
```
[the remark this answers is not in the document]
DriesSchaumont: Agreed

tests/test_xopen.py
```diff
@@ -69,6 +73,30 @@ def available_gzip_readers_and_writers():
...
+            ("pbzip2", PipedPBzip2Writer)
+        ]
+        if shutil.which(prog)
+    ]
+    return readers, writers
+
+
+PIPED_BZ2_READERS, PIPED_BZ2_WRITERS = available_bzip2_readers_and_writers()
+
+ALL_READERS_WITH_EXTENSION = list(zip(PIPED_GZIP_READERS, [".gz"])) + \
```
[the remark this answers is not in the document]
DriesSchaumont: Yes, indeed! Fixed.

DriesSchaumont:
@marcelm , When closing prematurely, pbzip2 does not exit with status code 15 when using sigterm, but with 1. My current solution is to catch the error message after termination from pbzip2  and check if it is OK to ignore the error or if the error was caused by another issue that caused exit status 1. Not a very clean solution. If you have a better idea, let me know!

src/xopen/__init__.py
```diff
@@ -330,18 +343,23 @@ def __iter__(self):
...
+                        allowed_message: Optional[Pattern] = None) -> None:
         """
-        Raise IOError if process is not running anymore and the exit code is
-        nonzero. If allow_sigterm is set and a SIGTERM exit code is
-        encountered, no error is raised.
+        Raise OSError if process is not running anymore and the exit code is
+        nonzero. If allowed_error_code is set and the exit value of the process is
+        equal to the value of allowed_error_code do not raise an error. If allowed_message
+        is set and does not match the output from stderr, OSError is raised even if the
+        exit code matches allowed_error_code.
```
marcelm, line 354: I believe the textual description doesn’t match the logic below.

src/xopen/__init__.py
```diff
@@ -330,18 +343,23 @@ def __iter__(self):
...
         """
         retcode = self.process.poll()
+        message = self._stderr.read().strip()
         if (
             retcode is not None and retcode != 0
-            and not (allow_sigterm and retcode == -signal.SIGTERM)
+            and not (allowed_error_code is not None and retcode == allowed_error_code)
+            and not (allowed_message is not None and re.search(allowed_message, message))
```
marcelm, line 361: I think the description in the docstring doesn’t match the code. 

This long list of conditions is getting hard to understand. One way to make it more readable is to negate the condition and to split it up into multiple ifs (De Morgan helps with this). That way, one could check the conditions under which we *don’t* need to raise an error one by one and return early. It would look sth. like this (note this is not what the final code should be, see below):
```python
if retcode is None:
    # process still running
    return
if retcode == 0:
    # process terminated successfully
    return
if retcode == allowed_error_code:
    # terminated with allowed error code
    return
if allowed_message is not None and allowed_message in message:
    # terminated with another exit code, but message is allowed
    return
...
raise OSError(...)
```
However, spelling this out, I think this is not what you intended because exit code 1 in this case would never raise an error. The docstring also doesn’t match the logic in the code.

marcelm: Ah, I think you can use the above as-is if you just don’t set an allowed error code for `pbzip2` (set it to `None`) *only* an allowed message.

src/xopen/__init__.py
```diff
@@ -441,6 +459,45 @@ def __init__(
...
+class PipedPBzip2Reader(PipedCompressionReader):
+    """
+    Open a pipe to pbzip2 for reading a bzipped file.
+    """
+    def __init__(self, path, mode: str = "r", threads: Optional[int] = None):
+        super().__init__(path, ["pbzip2"], mode, "-p", threads)
+
+    def _terminate(self):
+        """
+        Check if the process is running and terminate if necessairy.
```
marcelm, line 471: necessairy → necessary

src/xopen/__init__.py
```diff
@@ -308,20 +311,30 @@ def __repr__(self):
...
         )
 
-    def close(self) -> None:
-        if self.closed:
-            return
-        self.closed = True
+    def _terminate(self):
+        """
+        Check if the process is running and terminate if necessairy.
+        Return the expected error code and the expected error message
```
marcelm, line 317: "error code" should be "exit code" (also in a couple of other places below, in particular also in variable names, so `error_code` should be `exit_code`)

src/xopen/__init__.py
```diff
@@ -308,20 +311,30 @@ def __repr__(self):
             self._threads,
         )
 
-    def close(self) -> None:
-        if self.closed:
-            return
-        self.closed = True
+    def _terminate(self):
```
marcelm, line 314: Writing `_terminate` this way leads to some unnecessary code duplication down below when you need to override it in the `PipedPBzip2Writer` class. I also wouldn’t expect a function that is named `_terminate` to return an allowed error code and an allowed error message.

Perhaps you don’t need to add a `_terminate` method at all. You could just add class-level attributes to the `PipedCompressionWriter` like this:
```python
class PipedCompressionWriter(...):
    # This exit code is not interpreted as an error when terminating the process
    allowed_exit_code: Optional[int] = None
    # If this message is printed on stderr on terminating the process,
    # it is not interpreted as an error
    allowed_exit_message: Optional[str] = None
```
Then you set it appropriately in the subclasses and use it in `close()`.

tests/test_xopen.py
```diff
@@ -89,40 +119,72 @@ def fname(request):
...
+        yield
+
 
-    path = os.environ["PATH"]
-    os.environ["PATH"] = str(tmp_path)
-    yield
-    os.environ["PATH"] = path
+@pytest.fixture
+def lacking_pbzip2_permissions(tmp_path):
+    """
```
marcelm, line 158: Can you please move this docstring to `disable_binary` (and adjust it appropriately)?

tests/test_xopen.py
```diff
@@ -571,20 +657,23 @@ def test_pipedcompressionwriter_wrong_program():
 
 
 def test_compression_level(tmpdir, gzip_writer):
+    # Currenlty only the gzip writers handle compression levels.
```
marcelm, line 660: Currenlty → Currently

tests/test_xopen.py
```diff
@@ -442,17 +523,17 @@ def test_write_pigz_threads_no_isal(tmpdir, xopen_without_igzip):
         assert f.read() == 'hello'
 
 
-def test_read_gzip_no_threads():
```
marcelm, line 445: Did you remove this by accident? The function below tests writing, this one tests reading.

marcelm (commented):
Thanks again for the changes, I appreciate you’re taking the time to work on this. Sorry I have so many comments again ...

marcelm:
We’ll also need to install `pbzip2` as part of the GitHub Action workflow, but the tests should still pass even without `pbzip2` installed.

src/xopen/__init__.py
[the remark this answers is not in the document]
DriesSchaumont: Indeed, I was probably over-designing this. Adjusted

src/xopen/__init__.py
DriesSchaumont: @marcelm  Adjusted. I was wondering if we should make PipedCompressionWriter abstract and use abstract properties for this. This would force future expansions (e.g. parallel xz?) to check what the correct values for allowed_exit_code and allowed_exit_message should be and not use the defaults of `None`

tests/test_xopen.py
DriesSchaumont: Yes, I did. I restored it and expanded the test.

DriesSchaumont:
> We’ll also need to install `pbzip2` as part of the GitHub Action workflow, but the tests should still pass even without `pbzip2` installed.

I am not sure how to do this. But perhaps this does not need to be part of this PR?

DriesSchaumont:
> > We’ll also need to install `pbzip2` as part of the GitHub Action workflow, but the tests should still pass even without `pbzip2` installed.
> 
> I am not sure how to do this. But perhaps this does not need to be part of this PR?

I think I just learned how!

marcelm:
Just wanted to comment on installing pbzip2 during tests, but I see you added it already. We need to change this so that the tests get run both with and without `pigz` and `pbzip2` installed. Since these are optional requirements, we need to ensure that xopen works with and without them. But that does not need to be part of this PR.

marcelm:
I tried to find out where the tests hang on macOS. Here’s a CI run where added `-v` to the `pytest` invocation: https://github.com/pycompression/xopen/runs/2509169557?check_suite_focus=true

The tests hang in `test_reader_close` (that’s the next test that would run).

src/xopen/__init__.py
```diff
@@ -330,21 +337,34 @@ def __iter__(self):
...
         retcode = self.process.poll()
-        if (
-            retcode is not None and retcode != 0
-            and not (allow_sigterm and retcode == -signal.SIGTERM)
-        ):
-            message = self._stderr.read().strip()
-            self._file.close()
-            self._stderr.close()
-            raise OSError("{} (exit code {})".format(message, retcode))
+        message = self._stderr.read().strip()
```
marcelm, line 349: I think I found the cause of the hang on macOS: The `read()` in this line blocks. Previously, it would only be called if the process has already terminated. In this version, it is called even if the process is still running. (`retcode` is `None`).

You need to move this line to above `if check_allowed_code_and_message`

tests/test_xopen.py
```diff
@@ -89,40 +108,74 @@ def fname(request):
...
-    with open(large_gzip, 'a') as f:
-        f.truncate(os.stat(large_gzip).st_size - 10)
-    return large_gzip
+def create_large_file(tmpdir):
+    def _create_large_file(extension):
+        path = str(tmpdir.join(f"large{extension}"))
+        random_text = ''.join(random.choice('ABCDEFGHIJKLMNOPQRSTUVWXYZ\n') for _ in range(1024))
+        # Make the text a lot bigger in order to ensure that it is larger than the
+        # pipe buffer size.
+        random_text *= 1024
```
DriesSchaumont, line 164: @marcelm If I increase this value to 2048, `test_reader_close` hangs. I think this is a different issue to the incorrect placement of the `message = self._stderr.read().strip()`. This time, the test also hangs on linux. I think pbzip2 does not react to SIGTERM because the process is still running and it causes `self.process.wait()` to hang (I confirmed it by adding a timeout of 10 secongs with caused TimeoutExpired

DriesSchaumont: It does react to SIGKILL though.. which would require `self.process.kill()`

src/xopen/__init__.py
DriesSchaumont: I also figured that caused it so I moved it. However, I stumbled into a different issue. I started a new review to handle that one.

marcelm:
Just wanted to let you know that I’ll be away until Friday.

src/xopen/__init__.py
```diff
@@ -313,15 +321,20 @@ def close(self) -> None:
...
+        check_allowed_code_and_message = False
         if retcode is None:
             # still running
             self.process.terminate()
-            allow_sigterm = True
-        else:
-            allow_sigterm = False
-        self.process.wait()
+            check_allowed_code_and_message = True
+        try:
```
DriesSchaumont, line 329: I wonder if this is caused by https://github.com/pycompression/xopen/issues/16#issuecomment-5212402941

DriesSchaumont:
@marcelm I have fixed the issue with the files not closing correctly. It was caused by a deadlock where a process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data (see https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait). I fixed it by using a different thread to flush stdout so that SIGTERM  can reach the process. I think this is ready for another review.

src/xopen/__init__.py
marcelm: This would be a nice improvement. I’ll open an issue so we won’t forget (let’s do that in a separate PR).

.github/workflows/ci.yml
```diff
@@ -45,6 +45,13 @@ jobs:
     - name: Install isal
       if: matrix.with-isal && !startsWith(matrix.os, 'macos')
       run: sudo apt-get install isal libisal-dev
+    - name: Install pbzip2
```
rhpvorderman, line 48: Please do not use the bash statements. but use the yaml `if` clause instead. You may need to create two separate tasks (one for mac, one for linux) but that is okay.

src/xopen/__init__.py
```diff
@@ -243,11 +246,45 @@ def __next__(self):
         raise io.UnsupportedOperation('not readable')
 
 
+class DummyReader(Thread):
```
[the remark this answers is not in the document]
rhpvorderman: I am not a fan of using threads. These have a tendency to create problems faster than that they solve them. If `terminate()` blocks, send_signal can be used instead: https://docs.python.org/3.7/library/subprocess.html#subprocess.Popen.send_signal . And then use communicate or wait. Both have timeouts.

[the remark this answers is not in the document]
DriesSchaumont: One problem with `communicate()` is that it will close `self._stderr`, so if we want the error message we need to do `self._stderr.read()` _before_  `communicate()`. However, `read()` will block untill the EOF, meaning we need to process to be terminated for which we need `wait()` or `communicate()` first. Any suggestions?

.github/workflows/ci.yml
marcelm: I agree, but this is copied from how we did it in the "Install pigz" step just above it. I also suggest to install both `pigz` and `pbzip2` in the same step.

src/xopen/__init__.py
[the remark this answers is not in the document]
marcelm: `terminate()` just calls `send_signal(SIGTERM)` on Unix https://github.com/python/cpython/blob/115dea9e2602b96b63390f00cc880e90c433efa2/Lib/subprocess.py#L2092 so that should be enough.

@DriesSchaumont Can you say why you added the `DummyReader` class at all? The tests passed even before you added it. And the more I look at it, the less I understand why `wait()` would actually block. If `pbzip2` gets a SIGTERM while blocked on the output pipe, it still terminates correctly.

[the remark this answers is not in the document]
marcelm: `communicate()` returns what it reads from stdin and stderr (as a tuple), so you could pass it on to to `_raise_if_error`.

But as I said above, I don’t get why it’s necessary at all.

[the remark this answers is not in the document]
DriesSchaumont: @marcelm The problem is that pbzip2 is not killed by SIGTERM, which was causing the hanging tests (`test_reader_close`). The tests passed because I first implemented a solution where SIGKILL would be issued if SIGTERM failed to kill the process. I did not like this particular solution because if we use SIGKILL there is not way to distinguish between for example the kernel killing the process or xopen using  `kill()`. It also used a timeout, but different programs have different ways to handle SIGTERM so choosing a timer value seemed arbitrary...   I then noticed that pbzip2 does react to SIGTERM when the PIPE buffers are no longer full, which is why I implemented the seperate thread to flush it. 

As to _why_ pbzip2 is not closed by SIGTERM in this case, I am not sure... I will try to `communicate()` to see if I can avoid using custom thread.

src/xopen/__init__.py
```diff
@@ -243,11 +246,45 @@ def __next__(self):
...
+    """
+    Communication with a process may deadlock when using stdout=PIPE or
```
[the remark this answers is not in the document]
DriesSchaumont: I am closing this thread because it is outdated.

src/xopen/__init__.py
```diff
@@ -243,11 +246,45 @@ def __next__(self):
...
+    https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait
+    """
+    def __init__(self, pipe):
+        self._pipe = pipe
+        self._stop = Event()
+        super().__init__()
+
+    def run(self):
+        while not self._stop.is_set():
+            self._pipe.read()
```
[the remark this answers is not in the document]
DriesSchaumont: I am closing this thread because it is outdated.

src/xopen/__init__.py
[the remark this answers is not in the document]
DriesSchaumont: I removed the DummyReader in favor of using  `communicate()`. Closing this thread, please reopen if we should discuss this more.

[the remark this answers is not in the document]
marcelm: Ah, so are you on macOS? For me (on Linux), `pbzip2` does exit on SIGTERM.

But anyway, if `communicate()` works, then we have a solution.

.github/workflows/ci.yml
```diff
@@ -35,23 +35,19 @@ jobs:
           python-version: 3.7
           with-isal: true
     steps:
-    - name: Install pigz
+    - name: Install pigz and pbzip2 MacOS
+      if: startsWith(matrix.os, 'macos')
       run: >
-          if [[ ${{ startsWith(matrix.os, 'macos') }} = true ]]; then
             brew install pigz;
```
marcelm, line 41: Can this be written as `brew install pigz pbzip2` on one line?

.github/workflows/ci.yml
```diff
@@ -35,23 +35,19 @@ jobs:
...
-          else
+            brew install pbzip2
+    - name: Install pigz and pbzip2 Linux
+      if: startsWith(matrix.os, 'ubuntu')
+      run: >
             sudo apt-get install pigz;
```
marcelm, line 46: This can be written as `sude apt-get install pigz pbzip2`

rhpvorderman (approved):

rhpvorderman:
Looks good, nice and simple with lots of tests. A very good contribution. Thank you @DriesSchaumont !

.github/workflows/ci.yml
rhpvorderman: I made the commit just now.

marcelm:
I agree, this looks really good now, thanks again!

marcelm (approved):

PR #66, opened 2021-05-20
Wait for some output or process to exit after spawn.

### LINKED ISSUE #65: Tests failing after merging GH-63
Hi @marcelm ,

Unfortunatly, I noticed test are failing for merge commit https://github.com/pycompression/xopen/commit/aa45dcb12777ca196cc2ba414155ba9a8c9adbaa after enabling pbzip2.
I think the failure went unnoticed because commit https://github.com/pycompression/xopen/commit/4ff436d83c0ac00225b1d9ddbbf6544deb9f48f7 was made afterwards and this commit fails, but only for the code coverage. However, on https://github.com/pycompression/xopen/commit/aa45dcb12777ca196cc2ba414155ba9a8c9adbaa, test `test_nonexisting_file` fails: see https://github.com/pycompression/xopen/runs/2623845288

Seems like this test sometimes succeeds, while on other runs it fails. I am opening this issue to track our progress to resolve this.

DriesSchaumont:
closes #65

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -298,7 +298,7 @@ class PipedCompressionReader(Closing):
         assert self.process.stderr is not None
         self._stderr = io.TextIOWrapper(self.process.stderr)
         self.closed = False
-        self._wait_for_output_or_exit()
+        self._wait_for_output_or_process_exit()
         self._raise_if_error()
 
     def __repr__(self):
@@ -330,7 +330,7 @@ class PipedCompressionReader(Closing):
     def __next__(self) -> AnyStr:
         return self._file.__next__()
 
-    def _wait_for_output_or_exit(self):
+    def _wait_for_output_or_process_exit(self):
         """
         Wait for the process to procude at least some output, or has exited.
         """
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -345,7 +345,7 @@ class PipedCompressionReader(Closing):
             first_output = self.process.stdout.peek(1)  # type: ignore
             if first_output or self.process.poll() is not None:
                 break
-            time.sleep(0.1)
+            time.sleep(0.01)
 
     def _raise_if_error(self, check_allowed_code_and_message: bool = False,
                         stderr_message: bytes = b"") -> None:
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -28,6 +28,7 @@ import signal
 import pathlib
 import subprocess
 import tempfile
+import time
 from abc import ABC, abstractmethod
 from subprocess import Popen, PIPE, DEVNULL
 from typing import Optional, TextIO, AnyStr, IO, List, Set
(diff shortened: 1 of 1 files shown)


src/xopen/__init__.py
```diff
@@ -300,11 +300,15 @@ def __init__(
         # The program may crash due to a non-existing file, internal error etc.
         # In that case we need to check. However the 'time-to-crash' differs
         # between programs. Some crash faster than others.
-        # Therefore we peek the first character(s) of stdout. Peek will block
-        # until it gets at least the amount of characters specified or EOF.
-        # This way we ensure the program has at least decompressed some output,
-        # or stopped before we continue.
-        self.process.stdout.peek(1)  # type: ignore  # stdout is io.BufferedReader if set to PIPE.
+        # Therefore we peek the first character(s) of stdout. Peek will return at
+        # least one byte of data, unless at EOF in which case we should wait for the
```
marcelm, line 304: I think this sentence is a bit misleading because the docs say that `peek()` actually returns more or fewer bytes than requested https://docs.python.org/3.8/library/io.html#io.BufferedReader.peek . Because of that, I also wonder whether it is possible that the loop that you added below doesn’t terminate (that is, if the process keeps running but `peek()` keeps returning empty output).

marcelm: Ah, see this comment in the implementation of `BZ2File.peek` https://github.com/python/cpython/blob/bb3e0c240bc60fe08d332ff5955d54197f79751c/Lib/bz2.py#L145-L155:
> Relies on the undocumented fact that BufferedReader.peek() always returns at least one byte (except at EOF), independent of the value of n

src/xopen/__init__.py
```diff
@@ -300,11 +300,15 @@ def __init__(
...
+        # program to exit. This way we ensure the program has at least decompressed
+        # some output, or stopped before we continue.
+
+        # stdout is io.BufferedReader if set to PIPE
+        first_output = self.process.stdout.peek(1)  # type: ignore
+        while not first_output and self.process.poll() is None:
+            first_output = self.process.stdout.peek(1)  # type: ignore
```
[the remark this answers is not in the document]
marcelm: Does this need to be a loop at all?

src/xopen/__init__.py
DriesSchaumont: As far as I understood from this thread: https://bugs.python.org/issue5811 Seek should return at least one byte unless at EOF, but the documentation does not mention this.

src/xopen/__init__.py
```diff
@@ -300,11 +300,15 @@ def __init__(
         # The program may crash due to a non-existing file, internal error etc.
         # In that case we need to check. However the 'time-to-crash' differs
         # between programs. Some crash faster than others.
-        # Therefore we peek the first character(s) of stdout. Peek will block
-        # until it gets at least the amount of characters specified or EOF.
-        # This way we ensure the program has at least decompressed some output,
-        # or stopped before we continue.
-        self.process.stdout.peek(1)  # type: ignore  # stdout is io.BufferedReader if set to PIPE.
+        # Therefore we peek the first character(s) of stdout. Peek will return at
```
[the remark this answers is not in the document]
DriesSchaumont: Added.

DriesSchaumont:
Adjusted code for review and added error message.

src/xopen/__init__.py
```diff
@@ -336,6 +328,25 @@ def __iter__(self):
...
+        # least one byte of data, unless the buffer is empty or at EOF. If at EOF,
+        # we should wait for the program to exit. This way we ensure the program
+        # has at least decompressed some output, or stopped before we continue.
+
+        # stdout is io.BufferedReader if set to PIPE
+        while True:
+            first_output = self.process.stdout.peek(1)  # type: ignore
+            if first_output or self.process.poll() is not None:
+                break
+            time.sleep(0.1)
```
rhpvorderman, line 348: Can you make this `0.01`? When opening lots of files 0.1 can cause a very noticable slowdown. This also goes for 0.01 but it is 10 times faster and will be enough waiting time in 95% (estimated) of the cases.

rhpvorderman:
This looks good to me. I think the waiting time should be reduced to 0.01 but that's it.

src/xopen/__init__.py
DriesSchaumont: Yeah, 0.1 seems a bit excessive. Adjusted to 0.01.

marcelm (approved):

PR #74, opened 2021-09-20
Only use first error message when checking pbzip2 errors

### LINKED ISSUE #72: Test failure in test_truncated_with
This test appears to fail sporadically for the `bz2` case (and length 1024):
https://github.com/pycompression/xopen/runs/3636266596?check_suite_focus=true

https://github.com/pycompression/xopen/blob/7c2518bcedbd6a876d572339ac619384b80332d5/tests/test_xopen.py#L467-L473

I’m trying to reproduce it locally with `while pytest tests/test_xopen.py::test_truncated_with; do : ; done`, but haven’t succeeded.

rhpvorderman:
pbzip2 prints multiple error messages if terminated early, which wrongly triggers the allowed error messages in the checking codepath.

This should fix #72. Can you do the stress testing again @marcelm to check if this is the case?

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -486,6 +486,18 @@ class PipedPBzip2Reader(PipedCompressionReader):
     def __init__(self, path, mode: str = "r", threads: Optional[int] = None):
         super().__init__(path, ["pbzip2"], mode, "-p", threads)
 
+    def _raise_if_error(self, check_allowed_code_and_message: bool = False,
+                        stderr_message: bytes = b"") -> None:
+        # pbzip2 writes multiple messages. If a process is terminated after a
+        # reported error it will also display the terminate messages in stderr.
+        # This way we only check the original message.
+        if stderr_message:
+            # pbzip2 prints a newline first, so check second line.
+            messages = stderr_message.split(b"\n")
+            if len(messages) > 1:
+                stderr_message = messages[1]
+        super()._raise_if_error(check_allowed_code_and_message, stderr_message)
+
 
 class PipedPBzip2Writer(PipedCompressionWriter):
     """
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -369,7 +369,8 @@ class PipedCompressionReader(Closing):
             if retcode == self._allowed_exit_code:
                 # terminated with allowed exit code
                 return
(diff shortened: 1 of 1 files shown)


marcelm:
This crashes in the `_raise_if_error` in the superclass here:
```
        if not stderr_message:
            stderr_message = self.process.stderr.read()
```
```
  File "bug.py", line 11, in <module>
    f._file.read()
  File ".../xopen/src/xopen/__init__.py", line 135, in __exit__
    self.close()
  File ".../xopen/src/xopen/__init__.py", line 323, in close
    self._raise_if_error(check_allowed_code_and_message, stderr_message)
  File ".../xopen/src/xopen/__init__.py", line 499, in _raise_if_error
    super()._raise_if_error(check_allowed_code_and_message, stderr_message)
  File ".../xopen/src/xopen/__init__.py", line 378, in _raise_if_error
    stderr_message = self.process.stderr.read()
ValueError: read of closed file
```

marcelm:
For these errors, the message printed by pbzip2 is actually not preceded by a newline:
```
b"pbzip2: *ERROR: File ends unexpectedly!  Skipping...\n\n *Control-C or similar caught [sig=15], quitting...\nCouldn't signal signal QUIT to SignalHandlerThread [3]. Quitting prematurely!\n"
```
Then `msg.split(b"\n")[1]` becomes `b""`

marcelm:
Perhaps we could just be more strict and check whether the message is exactly `b"\n\nControl-C or similar caught [sig=15], quitting..."` (or perhaps `.startswith`), then we don’t even need to override `_raise_if_error`.

rhpvorderman:
Sorry, I accidently hit 'edit' instead of reply. A MorningError I am affraid. Hence the "edited by rhpvorderman" on your message.

That seems to be a good solution. This _allowed_error_message is for internal use only and should not be considered an API change. 

The exact error would be `\nControl-C or similar caught [sig=15], quitting...` (one `\n` less, because that was from the end of the File ends unexpectedly error.) And using startswith would be appropriate.

rhpvorderman:
> For these errors, the message printed by pbzip2 is actually not preceded by a newline:

Argh. Internal inconsistency! How can you account for that? Anyway it should be fixed now with the newest commit.

marcelm:
Very good, I’m no longer able to get my test script to crash.

I’ll publish version 1.2.0 later today since it’s been a while since the last release.

PR #75, opened 2021-09-21
Accomodate for /proc/sys/fs/pipe-max-size not being available due to permission errors

rhpvorderman:
I installed xopen on my phone (wanted to test something) and found that max pipe size was not available due to permission error. This fixes that issue.

@marcelm It would be nice if this lands in the next release.


marcelm:
Just saw this, I’ll wait. Will review shortly.

marcelm:
On your **phone**??? How did you do that?

marcelm:
Cool! (And that would finally be a good use case for the Bluetooth keyboard I bought on impulse a while ago)

PR #79, opened 2021-12-01
Do not install isal automatically on aarch64

rhpvorderman:
Decompressing is slower than native zlib with ISA-L.

I tested this on an aarch64 machine with ARM Cortex-A53 cores. (Allwinner A64 olimex olinuxino board). 

Compression is 2x faster. But decompression was about 30% slower. That is not a clear win. Unlike x86_64 where it decompresses 2x faster and compresses 5x faster.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,7 @@ package_dir =
     =src
 packages = find:
 install_requires =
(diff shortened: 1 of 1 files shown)


marcelm:
Thanks for testing this, LGTM!

PR #80, opened 2021-12-28
Add parameter to force a format

rhpvorderman:
I am currently working on a FASTQ filtering program for inclusion into [galaxy](https://galaxyproject.org/). 

One quirk of galaxy is that all datasets are written to `.dat` files. These `.dat` files are allowed to contain compressed data, as the metadata is stored in galaxy as well which specifies the exact format.

This means that fastq-filter which uses xopen will always default to `None` since `.dat` is not a known format. This results in people reaching their disk quota much faster as uncompressed files are about three times bigger.

I could of course workaround this by using `igzip.open` instead of xopen, but that would also lose me the flexibility of xopen in terms of:
- Threaded support
- Compression level automatically selecting the correct program (igzip vs pigz)

And it would require extra code to implement. Adding a simple `format` parameter to xopen was trivial, does not break backwards compatibility and will automatically choose the fastest way to write said format.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -704,7 +704,7 @@ def xopen(
     mode: str = "r",
     compresslevel: Optional[int] = None,
     threads: Optional[int] = None,
-    *,  # Keyword-only arguments below
+    *,
     format: Optional[str] = None,
 ) -> IO:
     """
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -703,6 +703,7 @@ def xopen(
     mode: str = "r",
     compresslevel: Optional[int] = None,
     threads: Optional[int] = None,
+    format: Optional[str] = None
 ) -> IO:
     """
     A replacement for the "open" function that can also read and write
@@ -732,6 +733,10 @@ def xopen(
     (parallel gzip) subprocess if possible. See PipedGzipWriter and PipedGzipReader.
 
     When threads = 0, no subprocess is used.
+
+    format overrides the autodetection of input and output formats. This can be
+    useful when compressed output needs to be written to a file without an
+    extension. Possible values are "gz", "xz" and "bz2".
     """
     if mode in ('r', 'w', 'a'):
         mode += 't'
@@ -742,7 +747,10 @@ def xopen(
     if filename == '-':
         return _open_stdin_or_out(mode)
 
-    detected_format = _detect_format_from_extension(filename)
+    if format not in (None, "gz", "xz", "bz2"):
+        raise ValueError(f"Format not supported: {format}. "
+                         f"Choose one of: 'gz', 'xz', 'bz2'")
+    detected_format = format or _detect_format_from_extension(filename)
     if detected_format is None and "w" not in mode:
         detected_format = _detect_format_from_content(filename)
 
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -587,8 +587,9 @@ def _open_bz2(filename, mode: str, threads: Optional[int]):
         except OSError:
             pass  # We try without threads.
 
-    # Ignore overzealous typing error from mypy.
-    # str is not an accepted mode type, it has to be Literal["rb"] etc.
+    # Ignore a TypeError that has been fixed in the typeshed.
+    # https://github.com/python/typeshed/pull/6722
+    # The ignore can be removed when a version of mypy with a synced typeshed is released.
     return bz2.open(filename, mode)  # type: ignore
 
 
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -703,7 +703,8 @@ def xopen(
     mode: str = "r",
     compresslevel: Optional[int] = None,
     threads: Optional[int] = None,
-    format: Optional[str] = None
+    *,  # Keyword-only arguments below
+    format: Optional[str] = None,
 ) -> IO:
     """
     A replacement for the "open" function that can also read and write
--- a/tests/test_xopen.py
+++ b/tests/test_xopen.py
@@ -723,7 +723,7 @@ def test_valid_compression_levels(writer, level, tmpdir):
     assert gzip.decompress(Path(test_file).read_bytes()) == b"test"
 
 
-def test_format_override(tmp_path):
+def test_override_output_format(tmp_path):
     test_file = tmp_path / "test_gzip_compressed"
     with xopen(test_file, mode="wb", format="gz") as f:
         f.write(b"test")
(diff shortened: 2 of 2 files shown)


rhpvorderman:
I ran into a crazy mypy error. See here: https://github.com/pycompression/xopen/runs/4648195093?check_suite_focus=true

Passing a `str` type as mode is not acceptable for bz2.open. It has to be one of `Literal["wb"]`, `Literal["rt"]` etc. This is extremely unhelpful typing. The bz2.open function already checks whether the modes are correct. As long as a `str` type is passed bz2 can do the checking correctly. Since the type system does not run at runtime, all this typing does is creating extra busywork. 

I will see if I can submit a bug upstream to correct this.

tests/test_xopen.py
```diff
@@ -724,3 +724,20 @@ def test_valid_compression_levels(writer, level, tmpdir):
     with writer(test_file, "wb", level) as handle:
         handle.write(b"test")
     assert gzip.decompress(Path(test_file).read_bytes()) == b"test"
+
+
+def test_format_override(tmpdir):
```
marcelm, line 729: It’s a bit inconsistent at the moment, but I’d like all tests to be converted to use `tmp_path` instead of `tmpdir` at some point. It would be great to use `tmp_path` in new code.

src/xopen/__init__.py
```diff
@@ -717,6 +721,10 @@ def xopen(
     (parallel gzip) subprocess if possible. See PipedGzipWriter and PipedGzipReader.
 
     When threads = 0, no subprocess is used.
+
+    format overrides the autodetection of input and output formats. This can be
```
marcelm, line 725: Being able to override the output format makes sense, but do you see a use case for overriding the input format? I guess it’s fine if only for symmetry ...

tests/test_xopen.py
```diff
@@ -724,3 +724,20 @@ def test_valid_compression_levels(writer, level, tmpdir):
...
+    assert test_contents.startswith(b"\x1f\x8b")  # Gzip magic
+    assert gzip.decompress(test_contents) == b"test"
+
+
+def test_format_override_unsupported_format(tmpdir):
+    test_file = tmpdir.join("test_fairy_format_compressed")
+    with pytest.raises(ValueError) as error:
+        xopen(test_file, mode="wb", format="fairy")
+    assert error.match("not supported")
+    assert error.match("fairy")
```
marcelm, line 743: It doesn’t hurt, but it’s inconsistent with the other tests, where you use the `.match` method without an `assert`.

src/xopen/__init__.py
```diff
@@ -688,6 +691,7 @@ def xopen(
     mode: str = "r",
     compresslevel: Optional[int] = None,
     threads: Optional[int] = None,
+    format: Optional[str] = None
```
marcelm, line 694: I believe it makes sense to turn this into a keyword-only argument (to force naming the argument)

marcelm (commented):
Adding this makes sense, I agree! (Although I think that Galaxy is doing a weird thing.)

src/xopen/__init__.py
rhpvorderman: Not really since we use the magic number to determine the filetype in case of input. (That is how Galaxy's `.dot` files are currently read correctly). 
One good reason to not seperate this between output and input is that it is much simpler, straightforward and least-surprising just to do it this way. And less code (less bugs...).

tests/test_xopen.py
rhpvorderman: Agreed. Done.

rhpvorderman:
Thank you for your timely review @marcelm  (I expected it after new year.) 
In full agreement with your comments and I have updated the code accordingly.

rhpvorderman:
Hmm, it seems the timeout gets easily triggered on windows.

marcelm:
> Hmm, it seems the timeout gets easily triggered on windows.

Ah, please increase it again (10 minutes should be fine). I reduced it when I was (ab)using CI for development because the job would sometimes hang.

src/xopen/__init__.py
```diff
@@ -703,6 +704,8 @@ def xopen(
     mode: str = "r",
     compresslevel: Optional[int] = None,
     threads: Optional[int] = None,
+    *,  # Keyword-only arguments below
```
marcelm, line 707: That comment should be removed

tests/test_xopen.py
```diff
@@ -721,3 +721,20 @@ def test_valid_compression_levels(writer, level, tmpdir):
     with writer(test_file, "wb", level) as handle:
         handle.write(b"test")
     assert gzip.decompress(Path(test_file).read_bytes()) == b"test"
+
+
+def test_format_override(tmp_path):
```
marcelm, line 726:
```suggestion
def test_override_output_format(tmp_path):
```

tests/test_xopen.py
```diff
@@ -721,3 +721,20 @@ def test_valid_compression_levels(writer, level, tmpdir):
...
+    test_file = tmp_path / "test_gzip_compressed"
+    with xopen(test_file, mode="wb", format="gz") as f:
+        f.write(b"test")
+    test_contents = test_file.read_bytes()
+    assert test_contents.startswith(b"\x1f\x8b")  # Gzip magic
+    assert gzip.decompress(test_contents) == b"test"
+
+
+def test_format_override_unsupported_format(tmp_path):
```
marcelm, line 735:
```suggestion
def test_override_output_format_unsupported_format(tmp_path):
```

src/xopen/__init__.py
marcelm: Ok, let’s keep it as it is.

tests/test_xopen.py
marcelm, line 726: Hm, I think also a test is needed that forces an incorrect input format, which then should raise an error (forcing gzip but reading uncompressed, for example).

src/xopen/__init__.py
```diff
@@ -732,6 +735,10 @@ def xopen(
     (parallel gzip) subprocess if possible. See PipedGzipWriter and PipedGzipReader.
 
     When threads = 0, no subprocess is used.
+
+    format overrides the autodetection of input and output formats. This can be
+    useful when compressed output needs to be written to a file without an
+    extension. Possible values are "gz", "xz" and "bz2".
```
marcelm, line 741: Is there a use case for forcing uncompressed input/output?

rhpvorderman: No. You just use `open` from builtins in that case. No need to bother with `xopen`.

marcelm:
You can cherry-pick 27e7dd9 (or merge #82 and rebase on top of main) to avoid the timeouts.

tests/test_xopen.py
rhpvorderman: I agree. Added it.

rhpvorderman:
Adressed the comments and rebased on top of main. Thank you for your review!

PR #81, opened 2021-12-31
Allow encoding, errors and newline parameters for text-mode

rhpvorderman:
This equalizes `xopen` with the `open` method from builtins. Also it is considered bad practice nowadays to open a text file without encoding, as that is not platform-independent behaviour. Encoding should always be explicitly passed. See https://docs.python.org/3.11/whatsnew/3.10.html#optional-encodingwarning-and-encoding-locale-option for more information.


rhpvorderman:
I already fixed the mypy error in #80 so I will just simply merge develop in once that is merged.

src/xopen/__init__.py
```diff
@@ -269,6 +277,10 @@ def __init__(
         mode: str = "r",
         threads_flag: Optional[str] = None,
         threads: Optional[int] = None,
+        *,
+        encoding=None,
```
marcelm, line 281: Should the default perhaps be `"utf-8"`? See:
https://docs.python.org/3.11/library/io.html#opt-in-encodingwarning
> If you are providing an API that uses open() or TextIOWrapper and passes encoding=None as a parameter, you can use text_encoding() so that callers of the API will emit an EncodingWarning if they don’t pass an encoding. However, please consider using UTF-8 by default (i.e. encoding="utf-8") for new APIs.

That should be mostly backwards compatible because we haven’t supported Windows so far, and the default on most Unix systems should be UTF-8 already.

If we leave it at `None`, we should use the `text_encoding()` function as suggested.

marcelm (commented):
Yes, this looks good, being closer to regular `open()` is a good thing. Just one comment.

src/xopen/__init__.py
rhpvorderman: `io.text_encoding()` Is a 3.10 only feature. I will set it to UTF-8.

rhpvorderman:
I added a small check to the `xopen` function. Setting the default encoding to `utf-8` caused crashes as the only default encoding allowed is `None` when the mode is binary. So now the xopen function does not pass the text mode args when the mode is "rb".

src/xopen/__init__.py
```diff
@@ -732,28 +777,36 @@ def xopen(
     (parallel gzip) subprocess if possible. See PipedGzipWriter and PipedGzipReader.
 
     When threads = 0, no subprocess is used.
+
+    encoding, errors and newline are used when opening in text mode. The parameters
+    have the same meaning as in the built-in open function.
```
marcelm, line 782: ... except that the default text encoding is UTF-8 on all platforms, not locale.getpreferredencoding().

rhpvorderman: Well-spotted. I should change that.

rhpvorderman:
@marcelm Done!
Also, I hope you had a nice holiday. It is good to be working with you again.

marcelm:
> Also, I hope you had a nice holiday. It is good to be working with you again.

Thanks! I’m happy to be back.

PR #83, opened 2022-01-13
Add seek() and tell() to PipedCompressionReader

marcelm:
These are just passed along to the underlying `_file` (the process` stdout).

This is needed by dnaio on Windows because `sys.stdin.buffer` on Windows is seekable, which triggers a different code path in `dnaio.singleend._detect_format_from_content` that relies on `seek()` and `tell()`.

Also, since `PipedCompressionReader.seekable()` already exists, we should also really have `seek()`.

~~Tests fail because coverage drops. I’ll see whether I find the energy to add some.~~ Test added

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -423,6 +423,12 @@ class PipedCompressionReader(Closing):
     def seekable(self) -> bool:
         return self._file.seekable()
 
+    def tell(self) -> int:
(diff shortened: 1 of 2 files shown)


rhpvorderman (approved):
Ready to merge!

marcelm:
Thanks for the review!

PR #84, opened 2022-01-14
Use Black

marcelm:
@rhpvorderman I’ve started to use Black in some of my projects and xopen would be another good candidate where it would make sense to use it. Although I don’t agree with all choices that Black makes, it is so much nicer not to have to talk about code formatting during PRs. (I saw some (minor) formatting issues in some of the recent PRs, but was already planning this PR, so I didn’t mention it.)

With this change, black-like formatting would be enforced during CI. To not have to remember to run `black` before each commit, pre commit hooks can be used, and it’s easiest to manage these with [https://pre-commit.com/](pre-commit). I added a pre-commit configuration file. To use it, run `pipx install pre-commit` and then `pre-commit install` in the repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
-        toxenv: [flake8, mypy]
+        toxenv: [black, flake8, mypy]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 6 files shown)


rhpvorderman:
I fully agree. Black may be less pretty at times, but it is mostly focused on creating as readable as possible git diffs. So that should make code review easier as well as keeping consistent formatting.

rhpvorderman (approved):

marcelm:
Great, then let’s do this.

PR #85, opened 2022-01-14
Use a bigger output buffer for writing to gzip and standard files

rhpvorderman:
Before:
```
Benchmark #1: python dnaio_read_and_write.py ~/test/big2.fastq ramdisk/output.fastq.gz
  Time (mean ± σ):      7.585 s ±  0.050 s    [User: 7.306 s, System: 0.278 s]
  Range (min … max):    7.512 s …  7.678 s    10 runs
```

After
```
Benchmark #1: python dnaio_read_and_write.py ~/test/big2.fastq ramdisk/output.fastq.gz
  Time (mean ± σ):      7.270 s ±  0.064 s    [User: 6.989 s, System: 0.279 s]
  Range (min … max):    7.129 s …  7.351 s    10 runs
```
Reading and writing dnaio without gzip is about 2 seconds. So the gzip compression time goes from 5.6 to 5.3. That is not too shabby.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -35,6 +35,7 @@ from typing import Optional, Union, TextIO, AnyStr, IO, List, Set
 
 from ._version import version as __version__
 
+# 128K buffer size also used by cat, pigz etc. It is faster than the 8K default.
 BUFFER_SIZE = max(io.DEFAULT_BUFFER_SIZE, 128 * 1024)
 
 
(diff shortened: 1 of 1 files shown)


rhpvorderman:
Also, this begs the question what default buffers python's gzip is using. Probably an 8k buffer as well.
Lately on python-isal I have updated all readers on the python-isal IgzipReader to be 128k. And finally finally, I got close to the speed of the C application igzip. Just 6% overhead. See: https://github.com/pycompression/python-isal/pull/106

rhpvorderman:
And probably we should checkout suboprocess.PIPE as well. That will probably not be a 128K buffer too.

marcelm:
Nice! Let me know if you think this justifies another release.

rhpvorderman:
I just tested the performance of updating Popen's bufsize argument with 128K. On reading the performance was the same. On writing the performance was very slightly better. (70ms saved). But it doesn't seem worth it to mess with it. I think the buffer sizes are fine now.

PR #88, opened 2022-01-17
Create somewhat smaller test files

marcelm:
This speeds up the tests, which take quite long at the moment.

### DIFF
--- a/tests/test_xopen.py
+++ b/tests/test_xopen.py
@@ -172,7 +172,7 @@ def create_large_file(tmp_path):
         )
         # Make the text a lot bigger in order to ensure that it is larger than the
         # pipe buffer size.
-        random_text *= 1024
+        random_text *= 2048
         with xopen(path, "w") as f:
             f.write(random_text)
         return path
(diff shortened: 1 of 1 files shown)


tests/test_xopen.py
```diff
@@ -163,16 +163,16 @@ def lacking_pbzip2_permissions(tmp_path):
...
+def create_large_file(tmp_path):
     def _create_large_file(extension):
         path = tmp_path / f"large{extension}"
         random_text = "".join(
             random.choice("ABCDEFGHIJKLMNOPQRSTUVWXYZ\n") for _ in range(1024)
         )
         # Make the text a lot bigger in order to ensure that it is larger than the
         # pipe buffer size.
-        random_text *= request.param
+        random_text *= 1024
```
rhpvorderman, line 175: The pipe buffer size is now exactly 1 MB since we always use the maximum available pipesize. Can the default be set to 2048 and the comment be slightly adjusted? 
Only processing 2048  and not 1024 and 4096 anymore should already help quite a lot with the speed.

rhpvorderman (changes requested):

rhpvorderman (approved):

rhpvorderman:
Good idea! I have just one nitpick.

tests/test_xopen.py
marcelm: Done!

PR #90, opened 2022-01-18
Speed up tests by two seconds

### LINKED ISSUE #89: pbzip2 waits one second when input is empty
Runninng `pytest --durations=10`, I noticed that `test_xopen_has_iter_method` with ext=`.bz2` takes one second to complete, which is surprising because it only checks whether an instance of `PipedPBzip2CompressionWriter` has the `__iter__` attribute.

It turns out that `pbzip2` has a delay of 1 second when piping empty input into it.
```
$ time echo -n|pbzip2 > out.bz2

real    0m1.004s
user    0m0.000s
sys     0m0.005s
```
For the tests, this can be fixed by just writing some data into the process – then there’s no delay.

marcelm:
Closes #89

Also some minor cleanups.

### DIFF
--- a/tests/test_xopen.py
+++ b/tests/test_xopen.py
@@ -446,10 +446,6 @@ def test_append_text(ext, tmp_path):
         assert appended == reference
 
 
-class TookTooLongError(Exception):
-    pass
-
-
 @pytest.mark.timeout(5)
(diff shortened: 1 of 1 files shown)


rhpvorderman:
LGTM

PR #93, opened 2022-02-25
Do not change default buffersize for builtin open

rhpvorderman:
I have recently found that having a small buffersize is most beneficial when writing large chunks. A larger buffersize is more beneficial when writing lots of small chunks. Buffersize is therefore application specific and the default of 8K seems to work best.

For dnaio it seems beneficial if the output buffer is 128K instead of 8K (small chunks) but when I tested `htspy` on writing uncompressed BAM (64K max chunks) it turned out to be slower. 
It is best to leave it to the application to determine this and otherwise choose the Python builtin default.


marcelm:
My faith in whoever set the defaults in the first place is restored ... ;-)

PR #96, opened 2022-03-07
Split tests into two files

marcelm:
The `test_xopen.py` file became a bit too big, so this PR splits the tests up into two files that contain different types of tests:
* `test_xopen.py` tests the `xopen.xopen` function itself
* `test_piped.py` tests the PipedCompression classes


marcelm:
With this change, there are two lines that are no longer covered by the tests: https://app.codecov.io/gh/pycompression/xopen/compare/96/changes#D1L104. I don’t have time to investigate, but I’m willing to ignore that, to be honest.

rhpvorderman:
Looks good to me. The missed lines are very rare error conditions anyway.

PR #97, opened 2022-03-08
Write reproducible gzip files in most cases

marcelm:
This
* enables `-n` for the external gzip processes except `python -m isal.igzip`, which doesn’t support it, yet.
* sets the equivalent options for `gzip.open` and `igzip.open` (which had to be replaced with `gzip.GzipFile` and `igzip.IGzipFile`

The test failure is due to `PipedPythonIsalWriter`.

To Do:
* [x] Add changelog entry

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,7 +38,7 @@ Python’s built-in ``gzip.open`` is used.
 
 When writing a gzip file (with any of the supported methods), it is written
 without timestamp and without the name of the original file (this corresponds
-to the ``gzip -n`` command-line option). Since `igzip adjusts its algorithm depending
+to the ``gzip --no-name`` command-line option). Since `igzip adjusts its algorithm depending
 on the CPU architecture <https://github.com/intel/isa-l/issues/140#issuecomment-634877966>`_,
 reproducible output is achieved only between runs on the same machine.
 
@@ -85,7 +85,7 @@ development version
 ~~~~~~~~~~~~~~~~~~~
 
 * Issue #94: When writing gzip files, the timestamp and name of the original
-  file is omitted (equivalent to using ``gzip -n`` on the command line).
+  file is omitted (equivalent to using ``gzip --no-name`` on the command line).
   This allows files to be written in a reproducible manner.
   (The bzip2 and xz compression methods do not store this information in the
   header and are therefore already reproducible.)
--- a/README.rst
+++ b/README.rst
@@ -38,7 +38,7 @@ Python’s built-in ``gzip.open`` is used.
 
 When writing a gzip file (with any of the supported methods), it is written
 without timestamp and without the name of the original file (this corresponds
-to the ``gzip --no-name`` command-line option). Since `igzip adjusts its algorithm depending
+to the ``gzip --no-name`` (or ``-n``) command-line option). Since `igzip adjusts its algorithm depending
 on the CPU architecture <https://github.com/intel/isa-l/issues/140#issuecomment-634877966>`_,
 reproducible output is achieved only between runs on the same machine.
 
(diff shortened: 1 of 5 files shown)


rhpvorderman:
GzipFile does not support "rt" mode, so that is something to take into account. I have put 'python -m isal.igzip' on my to do list: https://github.com/pycompression/python-isal/issues/110.

marcelm:
Yes, for text mode, I wrap the returned object in an `io.TextIOWrapper`. I just looked at the code for `gzip.open` and copied over the logic.

rhpvorderman:
Ah sorry, I didn't look good. Gave the code a good look now. Looks good to me. Just waiting for python -m isal.igzip then. And that is waiting for upstream ISA-L to release. So ETA unknown, unfortunately.

marcelm:
Yes, I agree, reproducibility on the same machine is good enough. We should point this out in the documentation.

rhpvorderman:
`python -m isal.igzip` now supports a `-n`/`--no-name` flag. The minimum required isal version should be bumped to `1.0.0` to support this.

rhpvorderman:
@marcelm, I just saw this due to e-mail notification. I think the unclosed file resource warning is what causes the issue. PyPy has some weird shennanigans going on in that department. On python-isal the PyPy tests failed because of a similar issue. When I run igzip.main() PyPy does not seem to close the files at the end of the main function. Hence the files are not completely written. On Cpython there are no problems. With Pypy I used `pypy -m isal.igzip` instead for testing and then it worked.

Still, I am not particularly fond of PyPy, but hopefully it is fixable.

marcelm:
Debugging this at the moment, I think it may actually be a bug in xopen that only surfaces when python-isal is not installed (as on PyPy). Will get back to you.

marcelm:
The problem (and this is even documented and I have even written a comment regarding this) is that `GzipFile` does not close the underlying file if a `fileobj` was passed to the constructor. However, to avoid storing the file name in the gzip file, we need to pass `filename=None` and instead use a `fileobj`.

CPython just hides the problem because garbage collection will call `close()` for us.

Now there’s a different failure, though, and I haven’t had the time to look into this.

marcelm:
The other error was just a good old logic bug, should be fixed now. Now there’s just a problem on macOS ...

rhpvorderman:
```
>>> timestamp, =  struct.unpack("<I", b'\\j\xeeb')
>>> datetime.date.fromtimestamp(timestamp)
datetime.date(2022, 8, 6)
```
Looks like the timestamp is not correctly set to 0 on MacOS. That is a bug with their gzip implementation.

rhpvorderman:
Found the error. The `-n` flag was forgotten on the PipedGzipWriter implementation.  I fixed the bug and also renamed all `-n` flags to `--no-name` because long options read just a little ieasier (IMHO). 
On the other hand, this does increase CLI parsing time, so we can no longer claim to be speed-focused ;-).

rhpvorderman:
> when python-isal is not installed (as on PyPy). 

Off-topic, but I readded PyPy support to python-isal. At first I copied a lot of code from the zlibmodule and adapted that. Including all the argument-clinic generated parser code.  PyPy wouldn't work with this. Lately I have written simple standard Python C-API parsers to get rid of a lot of byzantine generated code (https://github.com/pycompression/python-isal/pull/117). It turns out that this accidentally re-enabled the ability for PyPy to work with the extension. I only found this out when the conda package did the auto-update thing and compiled happily for PyPy. So I rushed to get 1.0.1 out of the door that enabled the wheelbuilding for PyPy again.

marcelm:
Thanks for fixing! I was a bit suprised this wasn’t caught by the tests, so I investigated a bit.

Apparently, `gzip` on Linux implicitly enables `-n`/`--no-name` when the input comes from stdin. The other tools (`igzip`, `pigz` and the macOS variant of `gzip`) leave out the file name (understandably), but still write the timestamp.

I added the `-n` in parentheses to the README so that it’s clear that `--no-name` is the same as `-n` (for people who may only know `-n`).

marcelm:
I’ll fix the coverage problem before merging.

PR #99, opened 2022-03-18
Make sure isal is only installed on CPython platforms

rhpvorderman:
PyPy will no longer be supported from version 1.0 onwards in python-isal.

PyPy + isal is slower than CPython + zlib, so if the PyPy users want extra compression speed it is better for them to move to a better version of Python first.


marcelm:
Not sure why codecov is reporting a decrease in coverage, but I’m going to ignore that.

PR #100, opened 2022-03-18
Drop Python 3.6 support

marcelm:
Python 3.6 is end-of-life and there have recently been new releases of
Python 3.7+ that fix security issues:
https://discuss.python.org/t/14353

Python 3.6 did not get these fixes, so it seems a good idea to stop using
and/or supporting it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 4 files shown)


marcelm:
I don’t understand why one of the codecov jobs shows decreased coverage, and for that matter, I don’t even understand why there *are* two codecov jobs, but I tend to trust the passing one. I’ll just merge this because I know @rhpvorderman is fine with dropping Python 3.6 in general.

PR #101, opened 2022-03-21
Add piped `xz`

### LINKED ISSUE #98: Support hooking into `xz` executable
In https://github.com/pycompression/xopen/pull/32#pullrequestreview-441546816, it was decided that `xz` compression would not get its own `PipedCompressionReader` and `PipedCompressionWriter` subclasses for the time being. I would like to take advantage of `xz`'s multithreaded compression option, but there's currently no way to do that within xopen. Do you think this would be reasonable enough to add?

fanninpm:
Fixes #98.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -654,15 +654,32 @@ class PipedPBzip2Writer(PipedCompressionWriter):
 
 class PipedXzReader(PipedCompressionReader):
     """
-    Open a pipe to xz for reading an xz-compressed file.
+    Open a pipe to xz for reading an xz-compressed file. A future
+    version of xz will be able to decompress using multiple cores.
+
+    (N.B. As of 21 March 2022, this feature is only implemented in xz's
+    master branch.)
     """
 
-    # TODO: implement threaded decompression once XZ releases it
     def __init__(
-        self, path, mode: str = "r", *, encoding="utf-8", errors=None, newline=None
+        self,
+        path,
+        mode: str = "r",
+        threads: Optional[int] = None,
+        *,
+        encoding="utf-8",
+        errors=None,
+        newline=None,
     ):
         super().__init__(
-            path, ["xz"], mode, encoding=encoding, errors=errors, newline=newline
+            path,
+            ["xz"],
+            mode,
+            "-T",
+            threads,
+            encoding=encoding,
+            errors=errors,
+            newline=newline,
         )
 
 
@@ -824,7 +841,7 @@ def _open_xz(filename, mode: str, threads: Optional[int], **text_mode_kwargs):
     if threads != 0:
         try:
             if "r" in mode:
-                return PipedXzReader(filename, mode, **text_mode_kwargs)
+                return PipedXzReader(filename, mode, threads, **text_mode_kwargs)
             else:
                 return PipedXzWriter(filename, mode, threads, **text_mode_kwargs)
         except OSError:
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -12,6 +12,8 @@ __all__ = [
     "PipedPigzWriter",
     "PipedPBzip2Reader",
     "PipedPBzip2Writer",
+    "PipedXzReader",
+    "PipedXzWriter",
     "PipedPythonIsalReader",
     "PipedPythonIsalWriter",
     "__version__",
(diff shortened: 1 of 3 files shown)


tests/test_piped.py
```diff
@@ -86,11 +88,24 @@ def available_bzip2_readers_and_writers():
 
 PIPED_BZIP2_READERS, PIPED_BZIP2_WRITERS = available_bzip2_readers_and_writers()
 
-ALL_READERS_WITH_EXTENSION = list(zip(PIPED_GZIP_READERS, cycle([".gz"]))) + list(
-    zip(PIPED_BZIP2_READERS, cycle([".bz2"]))
+
+def available_xz_readers_and_writers():
+    if shutil.which("xz"):
+        return [PipedXzReader], [PipedXzWriter]
+    return [], []
```
marcelm, line 95: The coverage decreases because this line is never called. It seems that `xz` is preinstalled on the machine that runs the CI job. You copied this from the equivalent function for `pbzip2`. It works for `pbzip2` because that is not preinstalled and we have two jobs, one that runs normally and one where we install `pbzip2`.

I think we can "cheat" a bit by writing it this way, which will give full coverage:
```
    result = [], []
    if shutil.which("xz"):
        result = [PipedXzReader], [PipedXzWriter]
    return result
```

However, I still want to test that the fallback to `lzma.open` works if the `xz` binary is not available. For this, can you please add an `xz`-version of `test_falls_back_to_bzip2_open`? You’ll also need to add a `lacking_xz_permission` fixture.

(We should clean these fixtures up a bit, but no need to do it in this PR.)

marcelm (commented):
Thanks! This looks quite good in my opinion.

src/xopen/__init__.py
```diff
@@ -650,6 +652,50 @@ def __init__(
         )
 
 
+class PipedXzReader(PipedCompressionReader):
+    """
+    Open a pipe to xz for reading an xz-compressed file.
+    """
+
+    # TODO: implement threaded decompression once XZ releases it
```
marcelm, line 660: Is this something that is planned from the xz side? It would be nicer to remove this comment and instead file an xopen issue to remind us.

marcelm:
This also needs an update to the README (in the introduction and changelog), but I’m happy to do this afterwards. Then I can do a little bit of restructuring at the same time.

src/xopen/__init__.py
```diff
@@ -650,6 +652,50 @@ def __init__(
...
+    def __init__(
+        self,
+        path,
+        mode: str = "wt",
+        threads: Optional[int] = None,
+        *,
+        encoding="utf-8",
+        errors=None,
+        newline=None,
+    ):
```
rhpvorderman, line 684: please add     `compresslevel: Optional[int] = None,`

To the parameters. This is quite relevant when compressing as the speed differences are quite substantial.

src/xopen/__init__.py
```diff
@@ -650,6 +652,50 @@ def __init__(
...
+        # Use default compression level for xz: 6
+        super().__init__(
+            path,
+            ["xz"],
+            mode,
+            6,
```
rhpvorderman, line 690: Please pass trough the compresslevel as highlighted above. A default of `None` will automatically set no flag and let xz choose the default compression level.

src/xopen/__init__.py
rhpvorderman: There is also `pixz` a parallel implementation of  `xz` . But it is less useful as xz already has threaded support.

From the XZ manual 

>   Threaded  decompression  hasn't  been  implemented yet.  It will
only work on files that contain multiple blocks with size infor‐
mation in block headers.  All files compressed in multi-threaded
mode meet  this  condition,  but  files  compressed  in  single-
threaded mode don't even if --block-size=size is used.

That means that adding a threads flag PipedXzReader is already valuable **now** as files that have been compressed in xz's threaded mode will decompress much faster when threaded mode is enabled for decompression.

rhpvorderman (changes requested):
I have some comments regarding the compression level and threads flags in addition.

src/xopen/__init__.py
fanninpm: > Is this something that is planned from the xz side?

I noticed that it was in xz's `master` branch (which hasn't been released yet), so I didn't know if it would be useful to pass the `-T` flag to the decompression stage. See:

- https://git.tukaani.org/?p=xz.git;a=commitdiff;h=4cce3e27f529af33e0e7749a8cbcec59954946b5
- https://git.tukaani.org/?p=xz.git;a=commitdiff;h=6c6da57ae2aa962aabde6892442227063d87e88c

rhpvorderman: Interesting. `pigz` also has "multithreaded" decompression, but Mark Adler stated that due to the format it was not really possible. So what he did is load off decompression, input processing and output processing to different threads, so that the decompression thread could use 100% of the CPU. 

The result is one thread that uses 100% and the whole process uses 130% (1.3 cores.) This 30% increase in CPU time leads to a 10% increase in throughput. 
For this reason xopen defaults to 1 thread for pigz whilst decompressing. 

It will be interesting to see how this dynamic works out for XZ. I get the feeling that when blocks have separate headers that speed will scale very well with number of threads. If not I expect more Pigz-like speed. If they manage to beat that expectation than it is pretty good indeed.

src/xopen/__init__.py
fanninpm: Implemented in 858ebba.

rhpvorderman (approved):

marcelm:
I was able to fix the test failure on Windows. I don’t quite know why test coverage decreases, but it seems unrelated (possibly some of the tests aren’t deterministic), and I’m therefore inclined to ignore it.

This looks good, thank you so much!

I’ll update the README and make a new release soon.

fanninpm:
Part of why test coverage decreased is because we aren't testing the xz `compressionlevel` bounds check.

marcelm:
Feel free to send a PR to fix that! (But you don’t have to.)

PR #104, opened 2022-03-22
Remove unused "type: ignore"

marcelm:
I noticed the comment regarding the TypeError fixed in typeshed. This can be removed now. I found the others with `mypy --warn-unused-ignores`.


rhpvorderman:
That is a nice trick to found unused ignores!

PR #106, opened 2022-04-01
Fix all EncodingWarnings

marcelm:
We run the tests with `PYTHONWARNDEFAULTENCODING=1`, which produces a couple of warnings in Python 3.10 (see [PEP 597](https://peps.python.org/pep-0597/)). This PR adds some encoding arguments to avoid those warnings. It also fixes an error where we were ignoring the encoding when opening stdout/stderr (with `"-"` specified as input or output file name).

The added `test_compression_writer_unusual_encoding` is just because I needed to convince myself that `PipedCompressionWriter` works correctly even for somewhat nonstandard encodings (it does, even without the changes in this PR).


rhpvorderman:
LGTM!

PR #108, opened 2022-04-12
Remove the special "isal" test environment from tox

marcelm:
This changes tox.ini to always build using pyproject.toml (`isolated_build = true`), which is more accurate as it also installs the dependencies as `pip install` would do.

Since this also installs all the Python-level dependencies including `isal`, the special tox environment named "isal" loses its meaning and is removed.

Instead, this configures the GitHub Actions workflow to test the library once with all external command-line tools installed and once without. (At least without those that are not pre-installed in the GitHub runner.)

Let’s see whether this changes the test coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,24 +28,23 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.7"]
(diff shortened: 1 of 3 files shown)


marcelm:
This actually improves coverage slightly, for example because we now have a test run where no command-line "igzip" is installed.

rhpvorderman:
Great work!

PR #110, opened 2022-08-10
Increase test coverage

marcelm:
A couple of commits that improve coverage a little bit.

In the "Run tests also without xz binary being available" commit, I wanted to get rid of the yellow color in line 94 in 
https://app.codecov.io/gh/pycompression/xopen/blob/09efe472ea488f38113fff8527a3725b035e882b/tests/test_piped.py. I initially used `sudo apt-get remove xz-utils`, but it turns out that there’s another `xz` binary in `/home/linuxbrew/.linuxbrew/bin/xz`.


rhpvorderman:
Looks good to me. It is always nice to add some additional corner-case tests in order to discover bugs before they happen.

marcelm:
So close to 100% now ...

PR #112, opened 2022-10-10
Add better type hints for xopen()

marcelm:
With this, xopen(..., mode="rb") correctly gets the return type "BinaryIO" and xopen(..., mode="rt") gets the return type TextIO.

Using this code from typeshed as a template:
https://github.com/python/typeshed/blob/fe3a34503cb54f30c932c329d40e615be14c1ddf/stdlib/builtins.pyi#L1432

To do: 
```
ImportError while loading conftest '/home/runner/work/xopen/xopen/tests/conftest.py'.
tests/conftest.py:6: in <module>
    from xopen import xopen
.tox/py/lib/python3.7/site-packages/xopen/__init__.py:1057: in <module>
    format: Optional[str] = ...,
E   TypeError: 'ABCMeta' object is not subscriptable
```

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -72,6 +72,9 @@ except OSError:  # Catches file not found and permission errors. Possible other 
     _MAX_PIPE_SIZE = None
 
 
+FilePath = Union[str, bytes, os.PathLike]
+
+
 def _available_cpu_count() -> int:
     """
     Number of available virtual or physical CPUs on this system
@@ -164,7 +167,7 @@ class PipedCompressionWriter(Closing):
 
     def __init__(
         self,
-        path: Union[str, bytes, os.PathLike],
+        path: FilePath,
         program_args: List[str],
         mode="wt",
         compresslevel: Optional[int] = None,
@@ -301,7 +304,7 @@ class PipedCompressionReader(Closing):
 
     def __init__(
         self,
-        path: Union[str, bytes, os.PathLike],
+        path: FilePath,
         program_args: List[Union[str, bytes]],
         mode: str = "r",
         threads_flag: Optional[str] = None,
@@ -1003,9 +1006,7 @@ def _open_reproducible_gzip(filename, mode, compresslevel):
     return gzip_file
 
 
-def _detect_format_from_content(
-    filename: Union[str, bytes, os.PathLike]
-) -> Optional[str]:
+def _detect_format_from_content(filename: FilePath) -> Optional[str]:
     """
     Attempts to detect file format from the content by reading the first
     6 bytes. Returns None if no format could be detected.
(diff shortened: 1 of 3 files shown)


rhpvorderman:
Very nice. `typing.Literal` is quite great. In my "try to make a better sam/bam reader project" it would be great to set certain types for certain tags.
Unfortunately 3.7 remains an issue until June of 2023.

If I might guess at the cause of the error: try to import Literal before the rest of the typing stuff. That way the typing_extensions import will not overwrite previously imported types.

marcelm:
Moving the imports around didn’t help unfortunately, but I think there’s a bug in Python 3.7 that lets it point at the wrong line when printing the error message. The actual problem seems to be the `os.PathLike[str]` part. Minimal example:
```
import os

def f(
    x: os.PathLike[str],
    y: int = 1,
): ...
```
Error message:
```
$ python -c 'import bug'
  File "<string>", line 1, in <module>
  File ".../bug.py", line 5, in <module>
    y: int = 1,
TypeError: 'ABCMeta' object is not subscriptable
```

marcelm:
Appears to be fixed!

src/xopen/__init__.py
```diff
@@ -1039,15 +1044,45 @@ def _detect_format_from_extension(filename: Union[str, bytes]) -> Optional[str]:
     return None
 
 
+@overload
+def xopen(
+    filename: Union[str, bytes, os.PathLike],
```
rhpvorderman, line 1049: What I usually do:
```Python3
FilePath = Union[str, bytes, os.Pathlike]
```
And then use FiePath everywhere for convenience, but YMMV.

rhpvorderman (approved):

src/xopen/__init__.py
marcelm: Done

src/xopen/__init__.py
```diff
@@ -1061,7 +1062,7 @@ def xopen(
 
 @overload
 def xopen(
```
rhpvorderman, line 1064: This breaks backward-compatibility for people who have filename as a keyword argument.

src/xopen/__init__.py
```diff
@@ -1061,7 +1062,7 @@ def xopen(
...
-    filename: Union[str, bytes, os.PathLike],
```
rhpvorderman: idem

src/xopen/__init__.py
```diff
@@ -1075,7 +1076,7 @@ def xopen(
 
 
 def xopen(  # noqa: C901  # The function is complex, but readable.
-    filename: Union[str, bytes, os.PathLike],
```
rhpvorderman, line 1043: idem

src/xopen/__init__.py
marcelm: Whoops, it was not my intention to sneak in something unrelated like this. Fixed now.

rhpvorderman:
Looks good to me!

marcelm:
Code coverage dropped, so I added a commit to exclude `@overload` from coverage reporting.

marcelm:
Thanks for the review!

PR #128, opened 2023-03-08
Write returns number of written bytes

### LINKED ISSUE #127: `write` should return the number of written bytes.
`write` should return the number of written bytes.

```python
import xopen
import gzip
import zstandard

fh_open = open("/tmp/open", "wb")
fh_xopen = xopen.xopen("/tmp/xopen.gz", "wb")
fh_gzip = gzip.open("/tmp/gzip.gz", "wb")
fh_zstandard = zstandard.open("/tmp/zstandard.zstd", "wb")

print(fh_open.write(b"texttext"))
# 8

print(fh_xopen.write(b"texttext"))
# None

print(fh_gzip.write(b"texttext"))
# 8

print(fh_zstdandard.write(b"texttext"))
# 8
```

rhpvorderman:
Maybe this issue would have been caught earlier if PipedCompressionWriter inherited from an abstract IO base class? It is something that should probably be done at some point. (Although I don't feel the urgency currently).

Fixes #127

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -268,8 +268,8 @@ class PipedCompressionWriter(Closing):
         process = Popen(program_args + extra_args, **kwargs)  # type: ignore
         return process
 
-    def write(self, arg: AnyStr) -> None:
-        self._file.write(arg)
+    def write(self, arg: AnyStr) -> int:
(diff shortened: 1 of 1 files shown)


marcelm:
Sorry for the delay, I kind of missed this

PR #130, opened 2023-10-02
Drop Python 3.7 and add Python 3.12

marcelm:


marcelm:
Will need to re-run this as soon as 3.12 is available on GitHub-hosted runners.

PR #132, opened 2023-10-18
Switch from setup.cfg to pyproject.toml

marcelm:
Maybe this can be merged before the 1.8.0 release. (I may need to rebase once #131 is merged.)


rhpvorderman:
I already fixed the mypy errors on my PR so I just merged it and rebased.

Mypy is getting really annoying. Typing is supposed to be a helpful tool that saves time. Not a nitpicking contest to waste developers time. I still think mypy is very beneficial, I just wish it wasn't so volatile and it wouldn't come up with new errors every few weeks.

marcelm:
Yeah, I agree somewhat. It would be nice if the way one is supposed to use typing in Python would stabilize ...

PR #133, opened 2023-10-20
Do not use piped igzip anymore

rhpvorderman:
Piped igzip is worse than python-isal's igzip_threaded module, but it is better than anything else.

But we have to consider the install base as well. My suspicion is that the only treason why people install igzip is because that is a dependency of xopen in conda and they install it automatically. If their platform supports igzip, it supports python-isal too. The python-isal conda package is build for the same architectures.

So the decision to remove it from the xopen decision tree is rather easy. What follows is the decision what to do about PipedIgzipReader and PipedIgzipWriter. I vote to put a DeprecationWarning in these classes and remove them one year after the release where they have been deprecated occurs. I doubt anyone is using them directly, and if so, we can always remove the deprecation later if people complain.

I want to see how xopen 1.8.0 does now it is released on Conda. There maybe some lingering bugs in the igzip_threaded module that can only be found with extensive use. Most bugs in python-isal are reported within one month after a release though. So after one or two months I will do the multithreaded thing to python-zlib-ng (using the powers of copy-paste) and integrate that in xopen. 

After that I feel that pigz should follow the same path as igzip. Pigz does not have a wide install base, it will be slower than python-zlib-ng. zlib-ng has a tremendously wide architecture support, there is no platform on conda that will not be able to run it. So as a result there will be little, if any users, that can install pigz, but not python-zlib-ng. Gzip will be universally installed on any linux system. It is even on alpine linux images (as part of busybox), so it does make sense to leave that in.

A niche case might be users that install cutadapt through debian packages. I see  that python3-xopen has a dependency on pigz. (But curiously enough, not on isal, even though that is available through the debian packaging system). 

I am just gonna leave this here as something to ponder about before the next release.

EDIT: And the PipedPythonIsalReader and PipedPythonIsalIgzipWriter should also be deprecated of course.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -1026,12 +1026,6 @@ def _open_external_gzip_reader(
 ):
     assert mode in ("rt", "rb")
     try:
-        return PipedIGzipReader(filename, mode, **text_mode_kwargs)
-    except (OSError, ValueError):
-        # No igzip installed or version does not support reading
-        # concatenated files.
-        pass
-    try:
         return PipedPigzReader(filename, mode, threads=threads, **text_mode_kwargs)
     except OSError:
         return PipedGzipReader(filename, mode, **text_mode_kwargs)
@@ -1041,11 +1035,6 @@ def _open_external_gzip_writer(
     filename, mode, compresslevel, threads, **text_mode_kwargs
 ):
     assert mode in ("wt", "wb", "at", "ab")
-    try:
-        return PipedIGzipWriter(filename, mode, compresslevel, **text_mode_kwargs)
-    except (OSError, ValueError):
-        # No igzip installed or compression level higher than 3
-        pass
     try:
         return PipedPigzWriter(
             filename, mode, compresslevel, threads=threads, **text_mode_kwargs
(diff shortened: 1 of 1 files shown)


marcelm:
Hi, I realized I never replied.
 
> So the decision to remove it from the xopen decision tree is rather easy.

I agree. Doing (de)compression in process should be preferred as it has fewer downsides and is more robust. Piping to or from an external process was only worth it because of the performance gain. I’ll be happy to see this simplified.

> What follows is the decision what to do about PipedIgzipReader and PipedIgzipWriter. I vote to put a DeprecationWarning in these classes and remove them one year after the release where they have been deprecated occurs. I doubt anyone is using them directly, and if so, we can always remove the deprecation later if people complain.

There are indeed very few usages within code hosted on GitHub, none of which seem to be relevant:
* [PipedIgzipReader](https://github.com/search?q=PipedIgzipReader&type=code)
* [PipedIgzipWriter](https://github.com/search?q=PipedIgzipWriter&type=code)

> I want to see how xopen 1.8.0 does now it is released on Conda. There maybe some lingering bugs in the igzip_threaded module that can only be found with extensive use. Most bugs in python-isal are reported within one month after a release though.

Due to my tardy reply, the month has now passed. Did you see an increase in bug reports? No Cutadapt user complained in any case.

> So after one or two months I will do the multithreaded thing to python-zlib-ng (using the powers of copy-paste) and integrate that in xopen.
> 
> After that I feel that pigz should follow the same path as igzip. Pigz does not have a wide install base, it will be slower than python-zlib-ng. zlib-ng has a tremendously wide architecture support, there is no platform on conda that will not be able to run it. So as a result there will be little, if any users, that can install pigz, but not python-zlib-ng. Gzip will be universally installed on any linux system. It is even on alpine linux images (as part of busybox), so it does make sense to leave that in.

Agreed to all of the above.

> A niche case might be users that install cutadapt through debian packages. I see that python3-xopen has a dependency on pigz. (But curiously enough, not on isal, even though that is available through the debian packaging system).

That’ll be up to the Debian maintainers to take care of.

> I am just gonna leave this here as something to ponder about before the next release.
> 
> EDIT: And the PipedPythonIsalReader and PipedPythonIsalIgzipWriter should also be deprecated of course.

Yes.

Shall we merge this PR as-is and do the rest in followup PRs?

rhpvorderman:
Thanks for your reply. Let's indeed merge it. I will put the python-zlib-ng thing on my to do list. It looks like the python-isal threaded implementation is very stable. I believe cutadapt with threads will use that one by default on most systems and I have not seen any complaints.

PR #137, opened 2024-01-12
Drop end of copyright year range

marcelm:
See   

* https://hynek.me/til/copyright-years/
* https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code

Since the commit message is going to be visible "forever" on the GitHub landing page for this project next to the file, I set it to just 'MIT license' on purpose.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2010-2021 xopen developers
+Copyright (c) 2010 The xopen developers
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


rhpvorderman:
I believe copyright takes 70 years to expire, so that practically means any software I write for work is the LUMC's copyright until well after I die (unless I become over a hundred, in which case I have probably other worries...). Doing extra bookkeeping to extend that copyright for certain parts of the software that were technically published later never seemed appealing to me. So I am on board.

PR #138, opened 2024-01-12
Set a default gzip compression level globally

rhpvorderman:
This simplifies the previous setup where the default was determined by the backend. 

I set the default to 1 for now. Most people will have python-isal installed, and this would have been the default for `None`-level compression for most users. I prefer promoting python-isal anywhere possible and as xopen states on the README that it chooses the most efficient way possible a level of 1 seems appropriate. 

I am also open to setting it to 5, however. I don't think it is that important. As most upstream applications have their own preferred default anyway.

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -1051,7 +1051,7 @@ def _open_gz(  # noqa: C901
                     mode,
                     compresslevel,
                     **text_mode_kwargs,
-                    threads=threads or max(_available_cpu_count(), 4),
+                    threads=1,
                 )
             except ValueError:  # Wrong compression level
                 pass
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -1060,41 +1060,42 @@ def _open_gz(  # noqa: C901
     assert mode in ("rt", "rb", "wt", "wb", "at", "ab")
     # With threads == 0 igzip_threaded defers to igzip.open, but that is not
     # desirable as a reproducible header is required.
-    if igzip_threaded and threads != 0:
-        try:
-            return igzip_threaded.open(  # type: ignore
-                filename,
-                mode,
-                isal_zlib.ISAL_DEFAULT_COMPRESSION  # type: ignore
-                if compresslevel is None
-                else compresslevel,
-                **text_mode_kwargs,
-                threads=1 if threads is None else threads,
-            )
-        except ValueError:  # Wrong compression level
-            pass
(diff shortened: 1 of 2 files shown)


rhpvorderman:
I forgot to mention, but this started as a refactor because I think the `_open_gz` function got a little messy. Setting a global default for compression level allowed removing a lot of code.

src/xopen/__init__.py
```diff
@@ -55,6 +54,9 @@
 # 128K buffer size also used by cat, pigz etc. It is faster than the 8K default.
 BUFFER_SIZE = max(io.DEFAULT_BUFFER_SIZE, 128 * 1024)
 
+# 1 is by far the fastest and most efficient level.
```
marcelm, line 57: I agree with your choice, but I’d prefer to avoid the hyperbole ("by far").

One could argue that "the fastest and most efficient level" would actually be 0 ... And level 9 would be the most efficient if you look at compression efficiency.

What I mean to say is: Choosing a compression level is an optimization problem with two criteria (speed and compression ratio). If you focus on one, you get the extremes ("no compression" and "infinite runtime"), so you need to choose a tradeoff that depends on the application. What we can do here in the library is to choose something sensible, but we shouldn’t call it "most efficient".

Can you just leave out the comment?

marcelm (commented):
Looks good! (Only one complaint about language ...)

I haven’t checked in detail whether the logic is still equivalent, but I’ll trust you and the tests.

src/xopen/__init__.py
rhpvorderman: That is a good idea. Unfortunately talking in hyperbole is one of my character flaws that is bound to pop up from time to time. Sorry about that. 

In this case I had switched the default around between 1, 5 and 6 a few times, because there are good reasons for all of these and the thought process got literary recoreded in the comment. I will remove it.

And that means yet another line less! I am quite happy with the refactoring so far.

rhpvorderman:
> I haven’t checked in detail whether the logic is still equivalent, but I’ll trust you and the tests.

I set a default threads at one point, but had to conclude that this changed the logic and changed it back. Also the reason to choose level 1 by default is that currently python-isal is chosen for compression when it is available. Setting it to 5 or 6 would change this behavior. So I think the logic is still equivalent, except that level 1 is chosen rather than level 2 (python-isal default) but that is OK (and more consistent).

src/xopen/__init__.py
rhpvorderman: Done and removed.

marcelm:
Thanks!

PR #139, opened 2024-01-15
Clean up README slightly and document the xopen function parameters

marcelm:


README.rst
```diff
@@ -72,6 +51,96 @@ and avoid using an external process::
...
+
+
+Backends
+--------
+
+Opening of gzip files is delegated to one of these programs or libraries:
+
+* `python-isal <https://github.com/pycompression/python-isal>`_.
+  Supports multiple threads and compression levels up to 3.
+* zlib-ng
```
rhpvorderman, line 131:
```
* `python-zlib-ng <https://github.com/pycompression/python-zlib-ng>`_. 
  Supports multiple threads and compression levels up to 9.
```

README.rst
```diff
@@ -72,6 +51,96 @@ and avoid using an external process::
...
+* `pigz <https://zlib.net/pigz/>`_ (a parallel version of ``gzip``)
+
```
rhpvorderman, line 133:
```
* `gzip <https://www.gnu.org/software/gzip/>`_. 
```

README.rst
```diff
@@ -72,6 +51,96 @@ and avoid using an external process::
...
+If multi-threaded compression or decompression is available,
+this parameter can be used to override the number of threads
+used. It is ignored otherwise.
+
+Set threads to 0 to force opening the file without using a subprocess.
+For some compression levels,
+compressed files are by default read or written
+using a pipe to a subprocess running an external tool such as
+``pbzip2`` or ``xz``.
+With *threads* set to 0, a normal function call is used instead.
```
rhpvorderman, line 121: The above is not entirely correct as not always a subprocess is used. Below is my attempt at the paragraph. It is a hard thing to properly explain in a few short sentences so please do not take this as a literal suggestion:

> Sets the number of threads spawned to open a compressed file. May be ignored if the backend does not support threads. 

> By default xopen tries to offload the (de)compression work load to other threads to free up the main Python thread for the application. This can either be done by using a subprocess to an external application or using a library that supports threads.

> Set threads to 0 to force xopen to use only the main Python thread.

README.rst
```diff
@@ -72,6 +51,96 @@ and avoid using an external process::
...
+file format and returns an open file-like object.
+
+When writing, the file format is chosen based on the file name extension:
+``.gz``, ``.bz2``, ``.xz``, ``.zst``. This can be overriden with ``format``.
+If the extension is not recognized, no compression is used.
+
+When reading and a file name extension is available, the format is detected
+from the extension.
+When reading and no file name extension is available,
+the format is detected from the contents.
```
rhpvorderman, line 82:
```
the format is detected from the `file signature <https://en.wikipedia.org/wiki/File_format#Magic_number>`_.
```

rhpvorderman (commented):
Nice cleanup! I have a few small suggestions.

README.rst
marcelm: Whoops, I wrote this while our broadband was updated and had no internet access. I had intended meant to add the link later.

marcelm: BTW, if you make these comments on the 'Files changed' tab, then you can use the 'Suggestions' button and I can just click to accept them.

See point 6 at https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request

README.rst
marcelm: I changed it to a variant of your suggestion. I agree it’s hard to explain. Maybe an extra section is needed in the documentation somewhere (later). As this paragraph is just the explanation of the parameter, I think it should be relatively short.

README.rst
rhpvorderman: Ah, that is good to know. Thanks!

rhpvorderman:
Looks good to me!

PR #140, opened 2024-01-15
Let PipedCompressionWriter/-Reader derive from IOBase

### LINKED ISSUE #129: Add all open methods to xopen.
Add all open methods to xopen.

If a method is not implementable, `UnsupportedOperation` exception could be generated like done in zstandard:

```python
import xopen
import gzip
import zstandard

fh_open = open("/tmp/open", "wb")
fh_xopen = xopen.xopen("/tmp/xopen.gz", "wb")
fh_gzip = gzip.open("/tmp/gzip.gz", "wb")
fh_zstandard = zstandard.open("/tmp/zstandard.zstd", "wb")

In [182]: fh_open.
                   close()      fileno()     mode         read()       readinto()   readlines()  tell()       write()     
                   closed       flush()      name         read1()      readinto1()  seek()       truncate()   writelines()
                   detach()     isatty()     raw          readable()   readline()   seekable()   writable()               

In [182]: fh_zstandard.
                        close()       flush()       read()        readinto()    seek()        truncate()    writelines() 
                        closed        isatty()      readable()    readline()    seekable()    writable()                 
                        fileno()      memory_size() readall()     readlines()   tell()        write()       

In [182]: fh_zstandard.readable()
Out[182]: False

In [183]: fh_zstandard.read()
UnsupportedOperation                      Traceback (most recent call last)
<ipython-input-183-6fbc0b8d75d3> in <module>
----> 1 fh_zstandard.read()

UnsupportedOperation: 

```

marcelm:
This is (probably) the proper thing to do for custom IO classes and also gives us a couple of methods for free, in particular `readlines()`.

This allows us to get rid of the `Closing` mixin (IOBase gives us the same functionality).

I had to also fix some warnings (due to `PYTHONDEVMODE=1`) caused by some objects being in a half-initialized stage when `close()` is called.

Closes #129

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -33,7 +33,6 @@ import pathlib
 import subprocess
 import tempfile
 import time
-from abc import ABC, abstractmethod
 from subprocess import Popen, PIPE, DEVNULL
 from typing import (
     Optional,
@@ -169,30 +168,7 @@ def _can_read_concatenated_gz(program: str) -> bool:
         os.remove(temp_path)
 
 
-class Closing(ABC):
-    """
-    Inherit from this class and implement a close() method to offer context
-    manager functionality.
-    """
-
-    def __enter__(self):
-        return self
-
-    def __exit__(self, *exc_info):
-        self.close()
-
-    def __del__(self):
-        try:
-            self.close()
-        except Exception:
-            pass
-
-    @abstractmethod
-    def close(self):
-        """Called when exiting the context manager"""
-
-
-class PipedCompressionWriter(Closing):
+class PipedCompressionWriter(io.IOBase):
     """
     Write Compressed files by running an external process and piping into it.
     """
@@ -228,7 +204,6 @@ class PipedCompressionWriter(Closing):
 
         # TODO use a context manager
         self.outfile = open(path, mode[0] + "b")
-        self.closed: bool = False
         self.name: str = str(os.fspath(path))
         self._mode: str = mode
         self._program_args: List[str] = program_args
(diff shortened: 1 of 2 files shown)


src/xopen/__init__.py
```diff
@@ -412,7 +391,7 @@ def close(self) -> None:
     def __iter__(self):
         return self
 
-    def __next__(self) -> Union[str, bytes]:
+    def __next__(self) -> Union[str, bytes]:  # type: ignore # incompatible with type in IOBase
```
rhpvorderman, line 394: This is something I ran into during my refactoring in #138 . The externally opened classes are slightly different in that they both support `t` and `b` modes. It would be much more in line with the rest of python if they are binary only, and a textwrapper is used around the pipedcompressionreader/writer objects. (Rather than interning a TextIOWrapper in the object as is done now). 

But that is something for later. I will make an issue about it.

rhpvorderman (approved):
Nice, this has been on the wishlist a long time!

src/xopen/__init__.py
marcelm: `TextIOWrapper` derives from `TextIOBase`, which derives from `IOBase`, so the problem that the return type changes in a derived class is still there. typeshed "solved" the issue the same way:
https://github.com/python/typeshed/blob/770724013de34af6f75fa444cdbb76d187b41875/stdlib/io.pyi#L145
I read a comment somewhere on the typeshed issue tracker stating that Python’s IO classes are just weird, and I agree.

It’s true we won’t have to deal with this when we make the Piped... classes work with binary data only. I agree and like the idea.

rhpvorderman: Good to know that the issue is not something that is necessarily wrong on xopen's end. Thank you for the explanation.

PR #145, opened 2024-01-29
Only open files in binary modes and use io.TextIOWrapper only at the highest level function.

### LINKED ISSUE #141: PipedCompressionReader and PipedCompressionWriter should be binary IO objects only.

rhpvorderman:
Fixes #141

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -38,7 +38,6 @@ from typing import (
     Optional,
     Union,
     TextIO,
-    AnyStr,
     IO,
     List,
     Set,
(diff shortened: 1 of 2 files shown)


marcelm:
Yes. Do you have time to make the release? (Update changelog, tag and push).

rhpvorderman:
Done. I see this branch is on my private fork for a reason. I will rebase the commit on main and start a PR with a branch on this repo.

PR #150, opened 2024-02-08
Support previously opened filehandles as input to xopen.open()

geertvandeweyer:
Minimal changes were applied to support passing open binary filehandles to xopen.xopen(). 

The tests were successful for all supported compression types, for both reading and writing, but I might have missed some of library dependencies.

Use case would be (and tested for) : 
- opening a non-compressed connection to a remote file using smart_open : 

`f_smart = smart_open('s3://mybucket/myfile.fastq.gz','wb', compression='disable')`

- then pass this handle to xopen to perform efficient (de)compression: 

`f = xopen.xopen(f_smart,mode='wb')`

I've noticed some mypy issues in the tox check. These are related to potentially conflicting object types. This could be fixed by explicitly renaming some variables to split flows depending on the input type (afaik). Let me know if you'd want this, or these issues can be ignored/whitelisted (?)

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -96,7 +96,7 @@ except (
     _MAX_PIPE_SIZE = None
 
 
-FilePath = Union[str, bytes, os.PathLike]
+FilePath = Union[str, bytes, os.PathLike, IO]
 
 
 def _available_cpu_count() -> int:
@@ -197,6 +197,9 @@ class PipedCompressionProgram(io.IOBase):
             raise ValueError(
                 f"Mode is '{mode}', but it must be 'r', 'rb', 'w', 'wb', 'a', or 'ab'"
             )
+        self._infile = path
+        if isinstance(path, io.IOBase):
+            path = path.name
         path = os.fspath(path)
         if isinstance(path, bytes) and sys.platform == "win32":
             path = path.decode()
@@ -240,7 +243,12 @@ class PipedCompressionProgram(io.IOBase):
         else:
             if compresslevel is not None:
                 self._program_args += ["-" + str(compresslevel)]
-            self.outfile = open(path, mode[0] + "b")
+            self._infile = path
+            # complex file handlers : pass through
+            if isinstance(self._infile, io.IOBase):
+                self.outfile = self._infile
+            else:
+                self.outfile = open(path, mode[0] + "b")
             try:
                 self.process = Popen(
                     self._program_args,
@@ -741,7 +749,11 @@ def _open_reproducible_gzip(filename, mode: str, compresslevel: int):
     # Neither gzip.open nor igzip.open have an mtime option, and they will
     # always write the file name, so we need to open the file separately
     # and pass it to gzip.GzipFile/igzip.IGzipFile.
-    binary_file = open(filename, mode=mode)
+    try:
+        binary_file = open(filename, mode=mode)
+    except TypeError:
+        # if an advanced filehandle is passed instead of filepath: pass it through
+        binary_file = filename
     kwargs = dict(
         fileobj=binary_file,
         filename="",
@@ -853,7 +865,7 @@ def xopen(  # noqa: C901  # The function is complex, but readable.
     A replacement for the "open" function that can also read and write
     compressed files transparently. The supported compression formats are gzip,
     bzip2, xz and zstandard. If the filename is '-', standard output (mode 'w') or
-    standard input (mode 'r') is returned.
+    standard input (mode 'r') is returned. Filenamem can be a string or a file object
 
     When writing, the file format is chosen based on the file name extension:
     - .gz uses gzip compression
@@ -898,8 +910,11 @@ def xopen(  # noqa: C901  # The function is complex, but readable.
     if mode not in ("rt", "rb", "wt", "wb", "at", "ab"):
         raise ValueError("Mode '{}' not supported".format(mode))
     binary_mode = mode[0] + "b"
+    # if file handle is passed: get name
+    file = filename
+    if isinstance(filename, io.IOBase):
+        filename = filename.name
     filename = os.fspath(filename)
-
     if format not in (None, "gz", "xz", "bz2", "zst"):
         raise ValueError(
             f"Format not supported: {format}. "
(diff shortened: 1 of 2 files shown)


src/xopen/__init__.py
```diff
@@ -96,7 +96,7 @@
     _MAX_PIPE_SIZE = None
 
 
-FilePath = Union[str, bytes, os.PathLike]
+FilePath = Union[str, bytes, os.PathLike, IO]
```
marcelm, line 99: This should be renamed to reflect the change in meaning. It should probably be `FileOrPath`.

src/xopen/__init__.py
```diff
@@ -898,8 +910,11 @@ def xopen(  # noqa: C901  # The function is complex, but readable.
     if mode not in ("rt", "rb", "wt", "wb", "at", "ab"):
         raise ValueError("Mode '{}' not supported".format(mode))
     binary_mode = mode[0] + "b"
+    # if file handle is passed: get name
+    file = filename
```
marcelm, line 914: I cannot add a review comment to the function signature above because it is too far away, so I’m writing it here: The `filename` parameter should be renamed to `file_or_path`.

tests/test_xopen.py
```diff
@@ -566,3 +566,67 @@ def test_xopen_zst_fails_when_zstandard_not_available(monkeypatch):
     with pytest.raises(ImportError):
         with xopen.xopen(TEST_DIR / "file.txt.zst", mode="rb", threads=0) as f:
             f.read()
+
+
+# tests for passing in filehandles
```
marcelm, line 571: This comment can be removed because it is redundant with the name of the test function

tests/test_xopen.py
```diff
@@ -566,3 +566,67 @@ def test_xopen_zst_fails_when_zstandard_not_available(monkeypatch):
...
+def test_passing_in_gz_filehandles_for_reading():
```
marcelm, line 572: The standard term to use would be ["file object"](https://docs.python.org/3/glossary.html#term-file-object) or file-like object instead of handle.

tests/test_xopen.py
```diff
@@ -566,3 +566,67 @@ def test_xopen_zst_fails_when_zstandard_not_available(monkeypatch):
...
+    first_line = CONTENT_LINES[0].encode("utf-8")
+    with open(TEST_DIR / "file.txt.gz", "rb") as fh:
+        with xopen(fh, "rb") as f:
+            assert f.readline() == first_line
+
+
+def test_passing_in_xz_filehandles_for_reading():
```
marcelm, line 579: Since this test is a copy of the one above with only the file extension changed, you can merge them into one. Just copy it how it is done above in other tests that use the `@pytest.mark.parametrize("ext", extensions)` decorator. This ensures that you test all the extensions that are supported (`.zst` is only supported optionally), and it also includes the empty extension (no compression), for which a test is missing at the moment.

tests/test_xopen.py
marcelm, line 572: Please also use the imperative mood (just  "pass", not "passing") for consistency with how the other tests are named.

marcelm (commented):
Hi and thanks for this! There PR #148 currently being worked on that also touches some of the code that this PR changes. Once #148 is merged, you’ll probably have to fix some merge conflicts. I’ve left a couple of comments already, but maybe you want to wait with addressing them until #148 is merged.

geertvandeweyer:
ok, thanks for the feedback. I didn't rename as this isn't my code :-) But I'll keep an eye on PR 148 and revise mine afterwards!

marcelm:
Hi @geertvandeweyer, PR #148 is now merged and, as expected, there are some conflicts. Let us know if you need help fixing them.

src/xopen/__init__.py
```diff
@@ -240,7 +243,12 @@ def __init__(  # noqa: C901
         else:
             if compresslevel is not None:
                 self._program_args += ["-" + str(compresslevel)]
-            self.outfile = open(path, mode[0] + "b")
```
rhpvorderman, line 243: The `w` mode case is handled appropriately. The `r` case cannot accept filehandles, as the external program needs to open files. Please make sure there are test cases for multiple threads for pbzip2 and xz (which will defer to an external program) and ensure these cases are handled properly. In this case the PipedCompressionProgram implementation should raise an error that is appropriately handled in the calling function.

rhpvorderman: I have some ideas  on how to make it work, but it will do it after your work is done :-).

geertvandeweyer: I've added a raise for now, as suggested.

geertvandeweyer:
I've addressed most comments I think. I'd welcome some feedback on a few issues though : 

- I'm hesitant to rename input arguments. It might break downstream applications. I've renamed filename to file, but kept support for filename for now
- I can't make the tests work correctly. the xopen_without_igzip doesn't seem to work for me.  Raising the ValueError on reading pre-opened gzip through pipedCompression, is not tested. 
- I didn't touch the test_piped.py tests. Help on implementing these is welcome.

rhpvorderman:
>  I'm hesitant to rename input arguments. It might break downstream applications. I've renamed filename to file, but kept support for filename for now

Currently the `xopen` function is the only function that should be considered as part of the stable API. A massive refactor took place in the last few weeks, where we purposefully broke all the other interfaces to get a massive saving of 40% less code. The other interfaces were always considered as internal use only, but this is made much more explicit now. As a consequence the next version of xopen will be 2.0.0. So it is theoretically possible to break a few things.

That said, I agree with you. The python stdlib `gzip.open` interface also supports opening filehandles and these parameter is called `filename` regardless. For the sake of consistency and not breaking users this should be kept as `filename`.

rhpvorderman:
Oh, I forgot to mention that I am willing to help with the tests as soon as you get it in a state that you feel it is ready. Please ping me when this is the case.

geertvandeweyer:
ok, thanks for helping with the tests! If the setup of allowing both file (default) and filename (with a warning) is fine, then the code is ready afaik.  I've tested what I can and it seems ok :-) 

If either filename **or** file should be supported, I can change the code again. Just let me know what you prefer.

rhpvorderman:
I think we should not support both. That is unnecessary complexity.

I see builtins.open uses `file`: https://docs.python.org/3/library/functions.html#open

However the compression modules, which offer an interface that is closer to xopen use `filename`:
https://docs.python.org/3/library/gzip.html#gzip.open
https://docs.python.org/3/library/bz2.html#bz2.open
https://docs.python.org/3/library/lzma.html#lzma.open

And then there is of course the backwards compatibility argument that puts me in the `filename` camp. 

Really, either of these names is fine, but since `filename` is used historically for xopen, we should stay with that. What are your thoughts @marcelm ?

marcelm:
`gzip.open` added support for passing in file objects in https://github.com/python/cpython/commit/68721019efb16ba8acad036c331a9a195d6f7da0. I assume the name of the parameter was left unchanged for backwards compatibility even back then. And `bz2.open` and `lzma.open` just copy what `gzip.open` does ...

It’s a little bit annoying that it’s called `filename`, but I’d be fine with leaving it as-is. That said, we wouldn’t be [breaking too many people’s code](https://github.com/search?q=xopen%28filename%3D&type=code). But there could be other code not on GitHub.

(Another thought: We could also turn it into a [positional-only parameter](https://docs.python.org/3/faq/programming.html#what-does-the-slash-in-the-parameter-list-of-a-function-mean).

I don’t have a strong opinion either way. I guess `filename` is the safest option.

geertvandeweyer:
Removed the file/filename in favor of filename only, as discussed.

geertvandeweyer:
thanks for picking up on this.  I'll leave it to you to finalize this then. Feel free to close this PR.

rhpvorderman:
Done. See #152

PR #151, opened 2024-02-09
Run black when testing through tox

marcelm:
I tested locally with `tox` and got confused because I was sure I had incorrectly formatted code. This is very fast du to `skip_install = True`, so we can run this first.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = flake8,mypy,py38,py39,py310,py311,py312,pypy3
+envlist = black,flake8,mypy,py38,py39,py310,py311,py312,pypy3
 isolated_build = True
 
 [testenv]


rhpvorderman:
Looks good to me.

As a side note: do you really have all these python versions in your active environment? In my solo projects I prefer using `py3` because that will select whatever is in my environment and not bother with the rest. I consent that you are the most thorough of us both though.

marcelm:
> As a side note: do you really have all these python versions in your active environment? 

Yes, I use [the Ubuntu deadsnakes PPA](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) to install all of these versions. But you’re right, that may be a bit much.

I’d be fine with changing the envlist to `black,flake8,mypy,py3` (note `py` also works as long as running `python` gives you Python 3) and then when I feel thorough, I could run `tox -e ALL`.

rhpvorderman:
No need to change the envlist. All non-installed versions are skipped anyway without throwing errors, so this works for everyone.

marcelm:
Ah, I didn’t know because I’ve always had all versions installed!

PR #152, opened 2024-03-11
xopen accepts filehandles

rhpvorderman:


rhpvorderman:
@geertvandeweyer It is finally done!
@marcelm Due to the enormous effort, the commits do not neatly follow how commits should ideally be done. Solving one issue created ResourceWarnings elsewhere and so forth. So I just committed when I got a step further on the ladder, in order to be able to deconstruct changes step by step. This actually saved me quite a few times. To remedy this, I suggest that this whole PR is squashed. Despite having spent 2+ days on tackling this the overall change in lines of code is quite small. I also had to fix python-isal and python-zlib-ng to enable this PR.

geertvandeweyer:
my apologies for the rabbit hole :-) 

and many thanks for this !

rhpvorderman:
@marcelm Friendly reminder ping

src/xopen/__init__.py
```diff
@@ -223,34 +224,37 @@ def __init__(  # noqa: C901
...
                 close_fds=close_fds,
             )  # type: ignore
+        except OSError:
+            if self.closefd:
+                self.fileobj.close()
+            raise
+        assert self.process.stdin is not None
+        if "r" in mode:
+            self.in_pipe = self.process.stdin
+            self.in_thread = threading.Thread(target=self._feed_pipe)
```
marcelm, line 252: Can you add a comment here or perhaps better to the `_feed_pipe` function why a thread is necessary/beneficial?

src/xopen/__init__.py
```diff
@@ -614,9 +652,42 @@ def _detect_format_from_extension(filename: Union[str, bytes]) -> Optional[str]:
...
+    if isinstance(file_or_path, (str, bytes)) or hasattr(file_or_path, "__fspath__"):
+        return open(os.fspath(file_or_path), binary_mode), True  # type: ignore
+    if isinstance(file_or_path, (io.BufferedReader, io.BufferedWriter)):
+        return file_or_path, False
+    if isinstance(file_or_path, io.TextIOWrapper):
+        return file_or_path.buffer, False
+    if isinstance(file_or_path, io.IOBase) and not hasattr(file_or_path, "encoding"):
+        # Text files have encoding attributes. This file is binary:
+        return file_or_path, False
+    else:
```
marcelm, line 671: For consistency, either do not use an `else` here an dedent the `raise` or use `elif` for the preceding `if`s.

src/xopen/__init__.py
```diff
@@ -614,9 +652,42 @@ def _detect_format_from_extension(filename: Union[str, bytes]) -> Optional[str]:
...
+        raise TypeError(
+            f"Unsupported type for {file_or_path}, "
```
marcelm, line 673: Hm, I wonder if this isn’t too restrictive. For the sake of duck typing, would it maybe be enough to not have most of the above cases and just check whether the object has a `read()` method instead?

src/xopen/__init__.py
```diff
@@ -614,9 +652,42 @@ def _detect_format_from_extension(filename: Union[str, bytes]) -> Optional[str]:
     return None
 
 
+def _file_or_path_to_binary_stream(
+    file_or_path: FileOrPath, binary_mode: str
+) -> Tuple[BinaryIO, bool]:
+    if binary_mode not in ("rb", "wb", "ab"):
+        raise AssertionError()
```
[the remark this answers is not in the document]
rhpvorderman: This might raise some type errors, becuase the "b" modes are not guaranteed. But I agree the code shouldn't pander too much to the type checker.

src/xopen/__init__.py
rhpvorderman: Agreed. I opted for "readinto" as that is done in dnaio IIRC for binary files.

rhpvorderman:
@marcelm All done. Thank you for the review!

tests/test_xopen.py
```diff
@@ -566,3 +567,44 @@ def test_xopen_zst_fails_when_zstandard_not_available(monkeypatch):
     with pytest.raises(ImportError):
         with xopen.xopen(TEST_DIR / "file.txt.zst", mode="rb", threads=0) as f:
             f.read()
+
+
+@pytest.mark.parametrize(["threads", "ext"], itertools.product((0, 1), extensions))
```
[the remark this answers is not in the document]
rhpvorderman: Cool, I didn't know this!

PR #156, opened 2024-03-27
Always set the maximum allowed window size for zstandard decompression

### LINKED ISSUE #155: xopen does not support .zst file compressed with Zstandard long memory mode
When compressing a file with long memory mode

`zstd -T4 -9 --force --rm --exclude-compressed --no-progress --long=31 file.txt`

Xopen cannot decode it
```
python3 -c 'import xopen; f = xopen.xopen("file.txt.zst"); f.readline()'
  File "<string>", line 1, in <module>
zstd.ZstdError: zstd decompress error: Frame requires too much memory for decoding
```

```
zstd -vl file.txt.zst
*** Zstandard CLI (64-bit) v1.5.5, by Yann Collet ***
tmpz/ip2ip_final.rpt.zst 
# Zstandard Frames: 1
DictID: 0
Window Size: 1.18 GiB (1264573175 B)
Compressed Size: 27.6 MiB (28914291 B)
Decompressed Size: 1.18 GiB (1264573175 B)
Ratio: 43.7352
Check: XXH64 d5d984dc
```
It seems the way to address this is [here in the lower level zstandard library](https://stackoverflow.com/questions/69270987/how-to-resolve-the-error-related-to-frame-used-in-zstandard-which-requires-too-m).

rhpvorderman:
Fixes #155


src/xopen/__init__.py
```diff
@@ -508,15 +508,26 @@ def _open_zst(
...
+    else:
+        max_window_bits = 31
     if threads != 0:
         try:
             # zstd can compress using multiple cores
+            program_args: Tuple[str, ...] = ("zstd",)
+            if "r" in mode:
+                # Only use --long=31 for decompression. Otherwise compression
+                # will always use a very long window size, which will result
+                # in very long compute times and the resulting archive not
```
marcelm, line 522: As I understand it (and my tests confirmed), a large window size doesn’t result in long compute times at compression time, only high memory usage

tests/only_zeroes.zst
marcelm, line 1: Can you document how this file was created, perhaps as a comment within the test using it?

tests/test_xopen.py
```diff
@@ -576,6 +576,18 @@ def test_xopen_zst_fails_when_zstandard_not_available(monkeypatch):
...
 
+@pytest.mark.parametrize("threads", (None, 0, 1))
+def test_xopen_zst_long_window_size(threads):
+    if threads == 0 and zstandard is None:
+        return
+    elif threads == 1 and not shutil.which("zstd"):
+        return
+    test_zst = Path(__file__).parent / "only_zeroes.zst"
+    with xopen(test_zst, "rb", threads=threads) as f:
+        data = f.read()
```
marcelm, line 587: This takes 0.5s on my system, which is impressively fast given that 2 GB of data are decompressed, but it’s still quite a long time for a test. I think it would be nicer to just read a couple of bytes from the start of the file. That is enough to detect whether the window size has been set correctly.

src/xopen/__init__.py
rhpvorderman: Considering a window size that does not fit in the CPU L1 or L2 cache will lead to very slow compute times. But the more important reason not to do it is that all the zstd files produced by xopen cannot be opened by zstd unless the user fiddles with the settings. That is very user unfriendly. Also --long=31 is very much unnessecary for most use cases.

So I will adjust the comment a bit.

tests/test_xopen.py
rhpvorderman: I only read 1024 bytes now and check if they are zero. This works wonders for the zstandard.open function. Unfortunately the command line application still takes a second to output its first data into stdout. Since there is no other way to test this properly I guess that is acceptable?

rhpvorderman:
@marcelm, I have addressed  the review comments.

tests/test_xopen.py
marcelm: Fine with me

marcelm (approved):

PR #158, opened 2024-03-27
Fix stdin bug

### LINKED ISSUE #157: xopen 2.0.0 does not read stdin fully
Originally logged as https://github.com/marcelm/cutadapt/issues/774 but reduced to an xopen only test case by @rhpvorderman :

```
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ pip list | grep xopen
xopen              2.0.0
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ wc -l ~/test/5millionreads_R1.fastq
20000000 /home/rhpvorderman/test/5millionreads_R1.fastq
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ cat ~/test/5millionreads_R1.fastq | python -c 'import xopen; f=xopen.xopen("/dev/stdin", "rt"); print(f.read())' | wc -l
19999956
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ cat ~/test/5millionreads_R1.fastq | python -c 'import xopen; f=xopen.xopen("-", "rt"); print(f.read())' | wc -l
19999956
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ pip install xopen==1.9.0 >/dev/null
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ cat ~/test/5millionreads_R1.fastq | python -c 'import xopen; f=xopen.xopen("-", "rt"); print(f.read())' | wc -l
20000001
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ cat ~/test/5millionreads_R1.fastq | python -c 'import xopen; f=xopen.xopen("/dev/stdin", "rt"); print(f.read())' | wc -l
20000001
(xopen) rhpvorderman@tuxminator:~/PycharmProjects/xopen$ 
```

The +1 is from print adding a new line.

rhpvorderman:
Fixes #157 

I also added a whole lot of tests to ensure piped reading never breaks again! (Hopefully)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -184,6 +184,13 @@ To ensure that you get the correct ``zstandard`` version, you can specify the ``
 Changelog
 ---------
 
+in-development
+~~~~~~~~~~~~~~~~~~~
++ #158: Fixed a bug where reading from stdin and other pipes would discard the
+  first bytes from the input.
++ #156: Zstd files compressed with the ``--long=31`` files can now be opened
+  without throwing errors.
+
 v2.0.0 (2024-03-26)
 ~~~~~~~~~~~~~~~~~~~
 
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -732,7 +732,7 @@ def _filepath_from_path_or_filelike(fileorpath: FileOrPath) -> str:
 def _file_is_a_socket_or_pipe(filepath):
     try:
         mode = os.stat(filepath).st_mode
-        return not (stat.S_ISFIFO(mode) or stat.S_ISSOCK(mode) or stat.S_ISPORT(mode))
+        return stat.S_ISFIFO(mode) or stat.S_ISSOCK(mode) or stat.S_ISPORT(mode)
     except (OSError, TypeError):  # Type error for unexpected types in stat.
         return False
 
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -10,6 +10,7 @@ __all__ = [
 
 import dataclasses
 import gzip
+import stat
 import sys
 import io
 import os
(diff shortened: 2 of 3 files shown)


marcelm (commented):
I’m not so sure about this fix. Explicitly checking all the cases in which we should not re-open the file in `_file_is_socket_or_pipe` feels fragile. Is it correct that the two `open()` calls occur in `_open_stdin_or_out`? Then an alternative solution would be to just not use `open()` there, but to just return the already open `sys.stdout.buffer` or `sys.stdin.buffer`. Would that work?

rhpvorderman:
I agree. When allowing file-like objects, my first design was to convert everything into a binary stream, then work from there. 
Unfortunately, this does not work properly with context managers. `gzip.open`, `bz2.open` etc all return an object that does not close the file, when the paramater passed is a file-like object rather than a path. Hence I need to pass a path to those functions if that was provided in the `xopen.xopen()` function. 

Multiple calls to the _open_as_binary_stream function solved this, but caused this problem. 

However I think I might alleviate this by using  a cache. That would be much less flaky than this current solution.

rhpvorderman:
A cache makes the code more messy. Pipes that are passed as files on linux (such as "/dev/stdin") need to be treated differently than regular files. The only way to do so is to have the stat code. 
The stat.S_ISREG  was used for this in xopen 1.9.0. Format detection was simply skipped for pipes. This was not fragile in that it only allowed regular files. (Are symlinks regular files?)

By doing it this way the format detection can be kept. I agree that it feels fragile in that not all edge cases are covered, but on the other hand checking for ports, sockets and pipes should be pretty much all the use cases.

rhpvorderman:
I have replaced all the `stat.S_ISFIFO` checks with a `not stat.S_ISREG` check. Symbolic links that point to regular files are also ISREG, so there is no problem there. This check has been in xopen for a very long time, so I trust this will work well.
I think with all the added tests in this PR we can be quite confident that the problem should be fixed long-term.

marcelm (approved):
Awesome with the extensive tests!

Issue #163, opened 2024-09-12
FR: Support for LZ4?


gcflymoto:
First of all thank you for such an excellent library which provides our engineering community with simple, transparent, performant, and pythonic library to handle all kinds of compressed and uncompressed files. It reduces the complexity by 10x of handling all these formats.

We've recently been deploying LZ4, especially since the latest release [1.10.0 supports multi-threaded compression and decompression](https://github.com/lz4/lz4/releases/tag/v1.10.0)

Unfortunately [https://github.com/python-lz4/python-lz4](https://github.com/python-lz4/python-lz4) development seems to have ceased.

Is there any possibility for adding python-zstandard style library? or support for invoking "lz4" CLI for now?

marcelm:
Thanks for the kind words!

My first thought is that adding LZ4 sounds reasonable. Are you interested in submitting a PR? A couple of thoughts in random order:
- Spawning `lz4` as a process would be easiest for now.
- Using python-lz4 as a library should also be needed, but the question is whether it should be a required or optional dependency. I don’t like that the lz4 wheel has 1.3 MB. This seems to be due to debugging info being included and could be reduced to around 100 kB. The Conda packages for lz4 itself plus the Python bindings (`lz4-c` and `lz4`) total <200 kB.
- python-lz4 apparently not being maintained is a bummer.
- Maybe the tests need some refactoring to avoid running too many of them?

@rhpvorderman What do you think?

marcelm:
Ah, great to hear it’s maintained. I’d be happy to have lz4 support, but don’t have time myself to work on this at the moment. So the answer at the moment is: PRs welcome!

rhpvorderman:
Yes, I am also in favour of adding LZ4. The python package not being ideal is not that bad as long as the conda package are good in terms of size. Those affect biocontainer sizes. 1.3MB is doable on most internet connections.

PR #168, opened 2025-03-07
Support for lz4 compression #163

gnzsnz:
- [X] changes on `__init__.py`
- [x] update `README.md`
- [x] update tests
- [x] update `ci.yml`
- [X] add `lz4` to dependencies
- [x] benchmark with and without `io.BufferedWriter` -> **remove** `io.BufferedWriter` for lz4
- [x] benchmark python lz4 vs lz4 CLI
- [x] set lz4 compression levels aligned with python lz4 [0-16], as CLi accepts those values without problem.


src/xopen/__init__.py
```diff
@@ -808,7 +853,7 @@ def xopen(  # noqa: C901
     compresslevel is the compression level for writing to gzip, xz and zst files.
     This parameter is ignored for the other compression formats.
     If set to None, a default depending on the format is used:
-    gzip: 6, xz: 6, zstd: 3.
```
rhpvorderman, line 811: Note to self: didn't we change the gzip level to 1?

pyproject.toml
```diff
@@ -5,30 +5,40 @@ build-backend = "setuptools.build_meta"
...
+dev = [
+    "pytest",
+    "pytest-timeout",
+    "tox",
+    "black",
+    "flake8",
+    "mypy",
+    "twine",
+    "setuptools_scm[toml]",
+]
```
rhpvorderman, line 39: :+1:

pyproject.toml
```diff
@@ -5,30 +5,40 @@ build-backend = "setuptools.build_meta"
...
 zstd = ["zstandard<1"]
+lz4 = ["lz4>=4.3.2"]
```
rhpvorderman, line 41: I propose making this non-optional. Yes, the linux binary is bigger than it should be, but it is just 1MB. The MacOS and Windows binaries are fine, and so are any binaries distributed on conda. But @marcelm needs to agree. 

I think we an address the issue upstream. The zstandard binding author could not be persuaded, but we might have more luck with the lz4 bindings author?

src/xopen/__init__.py
```diff
@@ -551,6 +558,39 @@ def _open_zst(
...
+                threads,
+                program_settings=_PROGRAM_SETTINGS["lz4"],
+            )
+        except OSError:
+            if lz4 is None:
+                # No fallback available
+                raise
+
+    if lz4 is None:
+        raise ImportError("lz4 module not available")
```
rhpvorderman, line 586: Looks good. What do we do when threads==0 but the lz4 program is available? Now it crashes, my inclination is to keep it that way if that is consistent with the behavior for the other formats. My idea is to always do what the user explicitly requested and crash if that is not possible. Although a case could be made that xopen should only take that as a suggestion, and prioritize getting the file open.

I just realized that this becomes irrelevant when lz4 is not an optional dependency anymore. Then this code can be simplified as well.

src/xopen/__init__.py
```diff
@@ -551,6 +558,39 @@ def _open_zst(
...
+    f = lz4.frame.LZ4FrameFile(filename, mode, compression_level=compresslevel)
+    if "r" in mode:
+        return f
+    # Buffer writes on lz4.open to mitigate overhead of small writes
```
rhpvorderman, line 590: For Gzip this overhead is present becase gzip is written in Python. Did you benchmark this to check if it made a differences for small writes?

rhpvorderman (changes requested):
Looks very good. My only question is to benchmark whether a bufferedwriter really adds value when returning an lz4 writable file. 

My other comment is that it is probably best to make lz4 non-optional, but that needs @marcelm 's blessing is well.

src/xopen/__init__.py
rhpvorderman: Yes we did.

rhpvorderman:
On debian 11 lz4 is not that picky. `-0` works and even `-1231` works (I assume anything in between works too). `--1` does not work. So technically you could use `range(0, sys.maxsize)` as an allowed range specifier. (Please do not use the tuple call in that case, python can check if something is in a range without the whole array being created).

src/xopen/__init__.py
rhpvorderman: Gzip and Bzip2 `write` calls are implemented in python. They have massive overhead. If the object you are writing to is implemented in C, that usually is not the case. I recommend benchmarking whether the BufferedWriter helps.

> or small writes we could use dictionaries for zstd and lz4, which should boost performance for small writes

Xopen does not support that. Gzip can also use dictionaries, but xopen does not provide the handles for that. That is more suited for low level libraries.

src/xopen/__init__.py
rhpvorderman: Yes. So in that case, lz4 CLI support can be slashed and the lz4 library be made a mandatory dependency. That should make the code much simpler. Supporting CLI programs was a mainstay of xopen back when there where no extra fast gzip libraries available and the fastest way to get things done was through a CLI program. 

But please benchmark first. Getting threaded performance with GIL dropping faster than a C CLI application is non-trivial. GIL-drop GIL-retain cycles have overhead and this can lead to behavior where the cores are not used to their full potential.

pyproject.toml
marcelm: I agree; I’m fine with making this non-optional. And we bring this up with the binding author(s).

rhpvorderman:
> Compression LZ4 python vs LZ4 CLI
> Python is faster up to 10MB (included), then CLI is better for big files at 100MB.

The CLI solution consists of the python process pushing data into the pipe and the LZ4 CLI processing that, effectively using 2 cores. This is using 1 additional thread.
The python lz4 is not given this benefit. What happens when python lz4 is opened with an additional thread (`threads=2` or whatever syntax they use)? That is the fair comparison.

> Decompression LZ4 python vs LZ4 CLI
> Mixed results here, CLI is faster for small files (1Kb), then python lz4 is faster. surprizingly 1Mb and 100Mb are really fast with python lz4, we might be hitting some block size sweet spot there. Overall for decompression python is better that CLI

The way you are benchmarking with 100 iterations is really sensitive to temporary moments when the CPU is busy doing other things, I think that explains the lacking 10 MB result. Except for 1kb, I see that for decompression Python LZ4 is always better. Given the large stddev on the 1KB result, I think that is from an outlier too.

My conclusion is that python lz4 is always faster. The decompression speed for lz4 is so enormously high (4GB/s and higher if the README is to be believed) that all we are doing is measuring the overhead of getting the data into a file. Using a pipe to another process is much more inefficient than doing it directly.


> There are benefits on not using io.BufferedWriter across different file sizes.

Great! Thanks for benchmarking that.

It looks like always going for python lz4 is the best option except for maybe compression. Can you compare the 2threads result with the piped result (using 1 thread on lz4) to have a more apples to apples comparison?

rhpvorderman:
> So, I don't think this test is possible. At least, I don't know how to do it. But i might be missing something.

> I agree with your conclusion, python LZ4 is the best for most of the scenarios. And when it's not the best, it's within the statistical error.

Ah, but if python-lz4 does not support a threaded mode, then it is better to use the pipedcompressionprogram class instead for scenarios where threading is requested. 

Python LZ4 drops the GIL, but if the threading library is not used to launch another thread, all the computation still happens on the same thread. So there is no speedup there. Xopen bypasses this with CLI programs. Python-isal is the exception here, because that contains quite some work to be able to escape the GIL and use true threading in Python. 

So I think the current PR is almost ready. Simply make sure python-lz4 is not optional but required and all the tests pass, and then I think it can be merged.

README.rst
```diff
@@ -27,6 +27,7 @@ Supported compression formats are:
 - bzip2 (``.bz2``)
 - xz (``.xz``)
 - Zstandard (``.zst``) (optional)
+- lz4 (``.lz4``) (optional)
```
rhpvorderman, line 30: Optional should be removed here btw.

README.rst
```diff
@@ -180,6 +181,20 @@ program or the Python ``zstandard`` package needs to be installed.
...
+
+For reading and writing lz4 (``.lz4``) files, either the ``lz4`` command-line
+program or the Python ``lz4`` package needs to be installed.
+
+* If the ``threads`` parameter to ``xopen()`` is ``None`` (the default) or any value greater than 0,
+  ``xopen`` uses an external ``lz4`` process.
+* If the above fails (because no ``lz4`` program is available) or if ``threads`` is 0,
+  the ``lz4`` package is used.
+
+To ensure that ``lz4`` is installed, you can specify the ``lz4`` extra for
```
rhpvorderman, line 195: This is also not true now lz4 is non-optional anymore.

rhpvorderman (changes requested):
I just checked the code if I could spot anything. But nothing yet. I will have to run this on my own PC to check why the tests are failing when I have the time.

cedricdonie:
Hi everyone, I came across this PR by chance and was interested in the discussion. If I understand it correctly, it would always be faster to use python-lz4 for decompression despite it being single-threaded. However, the current work in this PR seems to always call the piped compression program for decompression when threading is requested.

> Python LZ4 drops the GIL, but if the threading library is not used to launch another thread, all the computation still happens on the same thread. So there is no speedup there. Xopen bypasses this with CLI programs. Python-isal is the exception here, because that contains quite some work to be able to escape the GIL and use true threading in Python.

The above comment suggests that in theory, launching a subprocess should be faster because it can use a different process.

> > So, I don't think this test is possible. At least, I don't know how to do it. But i might be missing something.
> 
> > I agree with your conclusion, python LZ4 is the best for most of the scenarios. And when it's not the best, it's within the statistical error.
> 
> Ah, but if python-lz4 does not support a threaded mode, then it is better to use the pipedcompressionprogram class instead for scenarios where threading is requested.

Your interpretation seems to be that python is faster despite python using a single thread while the CLI used threads=auto. If we now use pipedcompressionprogram  when threading is requested, won't we slow down the decompression?

PS: I believe that sns.stripplots might be useful to analyze benchmarks with potential outliers. Here is a little notebook with an example: https://gist.github.com/cedricdonie/6785314197cee9a539b1936e22b2b982.
PPS: Thanks for the awesome library and for adding LZ4 🙂!

rhpvorderman:
> The above comment suggests that in theory, launching a subprocess should be faster because it can use a different process.

There are two definitions of "faster"

- Uses less wallclock time. This metric favours heavy threading.
- Uses less cpu time overall. This metric favours the least possible amount of overhead.

By using a subprocess, all the decompression is in one process. The python process only has to read input data from the pipe. It can then use the rest of the time to actually run the program you are interested in. This will decrease wall clocktime.

> Your interpretation seems to be that python is faster despite python using a single thread while the CLI used threads=auto. If we now use pipedcompressionprogram when threading is requested, won't we slow down the decompression?

LZ4 is a bit of a special case because the decompression is extremely fast. It could be that simply decompressing a block of data in LZ4 is much faster than incurring the overhead cost of a pipe while letting another program decompress it. So yes, that is a valid concern.

src/xopen/__init__.py
```diff
@@ -120,6 +126,7 @@ class _ProgramSettings:
     "zstd": _ProgramSettings(("zstd",), tuple(range(1, 20)), "-T"),
     "pigz": _ProgramSettings(("pigz", "--no-name"), tuple(range(0, 10)) + (11,), "-p"),
     "gzip": _ProgramSettings(("gzip", "--no-name"), tuple(range(1, 10))),
+    "lz4": _ProgramSettings(("lz4",), tuple(range(0, 17)), "-T"),
```
rhpvorderman, line 129: The bug is here. The "-T" flag has only been added very recently:
https://www.phoronix.com/news/LZ4-1.10-Multi-Threading

On most linux systems this will not be present yet. Debian 12 and Ubuntu 24.10 are still on 1.9.

rhpvorderman: Conda-forge is on 1.10. We have had this problem before with ISA-L, but the code base was much different. 

This can be fixed the easy way: simply remove the threads flag. Alternatively this can be fixed by some logic code that parses the version. That is a bit flaky though, if the author decides to change the version statement. Then there is another option of wrapping it in try ... except ... where first the threads flag is used and then no flag if an OSError is thrown.

Given LZ4 speed, I would go with the removing the threads flag. It is only really relevant for compressing with high levels of lz4 and that seems to be a not so typical use case. 

On my system, igzip level 1 (ISA-L optimized gzip) compresses just as fast as lz4 level 1 but with a much better compression ratio. lz4 level 9 reaches that compression ratio, but is much slower. I think lz4 will fare very well using a single thread compressing on level 1. For more compressed use cases gzip is a more relevant format. (Unless decompression speed is absolute key).

src/xopen/__init__.py
```diff
@@ -808,7 +850,7 @@ def xopen(  # noqa: C901
     compresslevel is the compression level for writing to gzip, xz and zst files.
     This parameter is ignored for the other compression formats.
     If set to None, a default depending on the format is used:
-    gzip: 6, xz: 6, zstd: 3.
+    gzip: 6, xz: 6, zstd: 3, lz4: 1.
```
marcelm, line 853: This doesn’t quite match that `XOPEN_DEFAULT_LZ4_COMPRESSION` is set to zero above.

tests/test_xopen.py
```diff
@@ -2,35 +2,37 @@
...
 import io
 import lzma
 import os
-from pathlib import Path
 import shutil
+import subprocess
+import sys
+import tempfile
+from contextlib import contextmanager
+from pathlib import Path
```
marcelm, line 15: I agree having sorted imports is nicer, but this should be done in a separate PR. (It’s ok this time, just leave it as it is.)

tests/test_xopen.py
```diff
@@ -2,35 +2,37 @@
...
 
-
 # TODO this is duplicated in test_piped.py
 TEST_DIR = Path(__file__).parent
 CONTENT_LINES = ["Testing, testing ...\n", "The second line.\n"]
 CONTENT = "".join(CONTENT_LINES)
 extensions = ["", ".gz", ".bz2", ".xz"]
 if shutil.which("zstd") or zstandard:
     extensions += [".zst"]
+if shutil.which("lz4") or lz4:
```
marcelm, line 34: With `lz4` support being non-optional, the `if` is no longer needed.

pyproject.toml
```diff
@@ -5,29 +5,39 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "xopen"
 authors = [
-    {name = "Marcel Martin", email = "marcel.martin@scilifelab.se"},
```
marcelm, line 8: Can you please undo these pure formatting changes? (Also below with `license`)

.gitignore
```diff
@@ -5,3 +5,10 @@ __pycache__/
 .tox
 venv/
 src/xopen/_version.py
+settings.json
```
marcelm, line 8: Please undo these changes as they are unrelated.

IMO, a `.gitignore` file should only contain project-specific patterns. There’s also a global ignore file at `~/.config/git/ignore` where one can put things like `.vscode` and `.DS_Store` that are specific to the tools used on the machine.

marcelm (approved):
Awesome, thanks! I finally got around to reviewing this PR. I just noted a couple of minor things that I think should be changed, but I will let @rhpvorderman make the final decision.

rhpvorderman:
python-lz4 is not tested and build for PyPy. So this is a problem with the upstream library or PyPy.

I am wondering what the best way is to work around this problem. Make lz4 optional again? Just drop the low quality bindings and only use the external program?

pyproject.toml
```diff
@@ -20,7 +20,8 @@ requires-python = ">=3.8"
 dynamic = ["version"]
 dependencies = [
     'isal>=1.6.1; platform.machine == "x86_64" or platform.machine == "AMD64" or platform.machine == "aarch64"',
-    'zlib-ng>=0.4.1; platform.machine == "x86_64" or platform.machine == "AMD64" or platform.machine == "aarch64"'
+    'zlib-ng>=0.4.1; platform.machine == "x86_64" or platform.machine == "AMD64" or platform.machine == "aarch64"',
+    'lz4>4.3.1'
```
marcelm:
```suggestion
    'lz4>4.3.1'; platform_python_implementation != "PyPy"',
```

marcelm:
> Personally, I think that lz4 as optional dependency is the most flexible approach. but is your call, please let me know.

Yes, but we could use an environment marker to require `lz4` only if we’re not on PyPy. It would look like this in `pyproject.toml`:
```
  'lz4>4.3.1'; platform_python_implementation != "PyPy"',
```
The code is a bit more complicated because we need to handle the case that `import lz4` fails, but that is already implemented AFAICS.

On PyPy, we then need to fall back to the `lz4` CLI.

A couple of observations regarding the CLI:
- Multithreading in `lz4` is only supported for compression, not for decompression.
- Older versions of `lz4` (such as v1.94 that comes with Ubuntu 24.04) do not support multithreading. They fail if one tries to pass the `-T` option.
- The newer versions use a default of `-T0`, which chooses the number of threads automatically (presumably the number of available cores).

So assuming that we do not want to try to detect which version of `lz4` is installed, we cannot use `-T`. We can therefore not actually control the number of threads used for compression; it will just depend on which version of `lz4` is installed.

Hm, what about this behavior:

Mode | Threads | What to use | Comment
-|-|-|-
Compress | 0 | python-lz4
Compress | !=0 | lz4 CLI | threads parameter is ignored, but if it is a newer binary, it will use all cores
Decompress | 0 | python-lz4
Decompress | != 0 | python-lz4 | threads parameter is ignored

And also fall back to lz4 CLI if `python-lz4` is not available.

So this is essentially `if lz4 is not None and (mode == 'rb' or (mode in ('ab', 'wb') and threads == 0)): use Python bindings; else: use lz4 CLI`

What do you think?

rhpvorderman:
> And also fall back to lz4 CLI if python-lz4 is not available.

Also if threads was explicitly set to 0? I am in dubio about that one. Crashing because a single-threaded option can not be provided leads to the user not being able to open the file. Defaulting to the CLI will use more threads than expected, which can cause issues on cluster environments.

marcelm:
> Also if threads was explicitly set to 0? I am in dubio about that one. Crashing because a single-threaded option can not be provided leads to the user not being able to open the file. Defaulting to the CLI will use more threads than expected, which can cause issues on cluster environments.

Hm, you’re right.

In general, I would say that as an xopen user, my main interest is in being able to open files that may be compressed, not caring so much about the details. Sure I can specify compression level and threads, but this is more on a “best effort” level. For example, if we open a bz2 file with threads > 0 but pbzip2 is not available, we open it using `bz2.open` anyway (singlethreaded).

Applied to lz4, I would say we should not fail even when the bindings aren’t available and threads == 0.

How about this instead of version detection (which I’d like to avoid): We try to run the `lz4` CLI with the appropriate `-T` option first, and if that fails, we just re-try without the option.

Pseudocode
```
if lz4 is not None and (mode == 'rb' or (mode in ('ab', 'wb') and threads == 0)):
    # use Python bindings
else:
    try:
        run lz4 with option -T max(threads, 1)
    except ...:
        run lz4 without option -T
```

rhpvorderman:
Fully agree on all your points. That seems to be the best option.

gnzsnz:
**Status**

- [x] lz4 dependency, but not for pypy on pyproject. `'lz4>4.3.1'; platform_python_implementation != "PyPy"'`
- [x] lz4 dependency management. The previous point makes tox fail for pypy. we need to consider a logic where lz4 is still treated as an optional dependency. this was part of the initial commit. i need search the commit history and apply it.
- [x] logic to maximize changes of lz4 success https://github.com/pycompression/xopen/pull/168#issuecomment-2771944043
- [ ] there is a new error on `test_xopen.py::test_pass_bytesio_for_reading_and_writing` for params [lz4-1]. error `io.UnsupportedOperation: fileno`. After my initial analysis is clear that BytesIO will not work for process pipes, as there is no functional `fileno` method. My understanding is that this is testing threads on a python module. for lz4 it works with `threads == 0` but when `threads == 1` it tries to use LZ4 CLI which fails. i can add a check on the code or maybe an xfail marker. Need to look into it.

let me know your comments

gnzsnz:
- [x] merged latest changes form origin/main into `lz4` branch
- [x] issues with pypy3 solved. lz4 is treated as an optional dependency.
- [ ] `test_pass_bytesio_for_reading_and_writing` still failing. it's weird tox tests are working for all compression formats but lz4. I would expect the same error for all formats.

I could use some help on the `BytesIO` issue.

Reproduction steps:

```python
# gzip test
from io import BytesIO
from xopen import xopen
filelike = BytesIO()
msg = b"my test message"
with xopen(filelike,mode="wb",format="gz",threads=1) as f:
    f.write(msg)
filelike.seek(0)
with xopen(filelike,mode="rb",format="gz",threads=1) as f:
    read_msg = f.read()
assert msg == read_msg

# lz4 threads = 0 (lz4 module)
filelike = BytesIO()
msg = b"my test message"
with xopen(filelike,mode="wb",format="lz4",threads=0) as f:
    f.write(msg)
filelike.seek(0)
with xopen(filelike,mode="rb",format="lz4",threads=0) as f:
    read_msg = f.read()
assert msg == read_msg

# lz4 threads =1 (cli lz4 command)
filelike = BytesIO()
msg = b"my test message"
with xopen(filelike,mode="wb",format="lz4",threads=1) as f:
    f.write(msg)
filelike.seek(0)
with xopen(filelike,mode="rb",format="lz4",threads=1) as f:
    read_msg = f.read()
assert msg == read_msg
# error
UnsupportedOperation                      Traceback (most recent call last)
File ~/Documents/dev/xopen/src/xopen/__init__.py:578, in _open_lz4(filename, mode, compresslevel, threads)
    577 try:
--> 578     return _PipedCompressionProgram(
    579         filename,
    580         mode,
    581         compresslevel,
    582         threads,
    583         program_settings=_PROGRAM_SETTINGS["lz4"],
    584     )
    585 except OSError:

File ~/Documents/dev/xopen/src/xopen/__init__.py:227, in _PipedCompressionProgram.__init__(self, filename, mode, compresslevel, threads, program_settings)
    225 self._threads = threads
--> 227 self._open_process()

File ~/Documents/dev/xopen/src/xopen/__init__.py:252, in _PipedCompressionProgram._open_process(self)
    251 try:
--> 252     self.process = subprocess.Popen(
    253         self._program_args,
    254         stderr=self._stderr,
    255         stdout=stdout,
    256         stdin=subprocess.PIPE,
    257         close_fds=close_fds,
    258     )  # type: ignore
    259 except OSError:

File ~/.pyenv/versions/3.12.9/lib/python3.12/subprocess.py:994, in Popen.__init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags, restore_signals, start_new_session, pass_fds, user, group, extra_groups, encoding, errors, text, umask, pipesize, process_group)
    977 # Input and output objects. The general principle is like
    978 # this:
    979 #
   (...)    989 # are -1 when not using PIPEs. The child objects are -1
    990 # when not redirecting.
    992 (p2cread, p2cwrite,
    993  c2pread, c2pwrite,
--> 994  errread, errwrite) = self._get_handles(stdin, stdout, stderr)
    996 # From here on, raising exceptions may cause file descriptor leakage
    997 
    998 # We wrap OS handles *before* launching the child, otherwise a
    999 # quickly terminating child could make our fds unwrappable
   1000 # (see #8458).

File ~/.pyenv/versions/3.12.9/lib/python3.12/subprocess.py:1730, in Popen._get_handles(self, stdin, stdout, stderr)
   1728 else:
   1729     # Assuming file-like object
-> 1730     c2pwrite = stdout.fileno()
   1732 if stderr is None:

UnsupportedOperation: fileno

During handling of the above exception, another exception occurred:

UnsupportedOperation                      Traceback (most recent call last)
Cell In[43], line 3
      1 filelike = BytesIO()
      2 msg = b"my test message"
----> 3 with xopen(filelike,mode="wb",format="lz4",threads=1) as f:
      4     f.write(msg)
      5 filelike.seek(0)

File ~/Documents/dev/xopen/src/xopen/__init__.py:913, in xopen(filename, mode, compresslevel, threads, encoding, errors, newline, format)
    911     opened_file = _open_zst(filename, binary_mode, compresslevel, threads)
    912 elif detected_format == "lz4":
--> 913     opened_file = _open_lz4(filename, binary_mode, compresslevel, threads)
    914 else:
    915     opened_file, _ = _file_or_path_to_binary_stream(filename, binary_mode)

File ~/Documents/dev/xopen/src/xopen/__init__.py:588, in _open_lz4(filename, mode, compresslevel, threads)
    586 _program_settings = _PROGRAM_SETTINGS["lz4"]
    587 _program_settings.threads_flag = None
--> 588 return _PipedCompressionProgram(
    589     filename,
    590     mode,
    591     compresslevel,
    592     threads,
    593     program_settings=_program_settings,
    594 )

File ~/Documents/dev/xopen/src/xopen/__init__.py:227, in _PipedCompressionProgram.__init__(self, filename, mode, compresslevel, threads, program_settings)
    224         threads = min(_available_cpu_count(), 4)
    225 self._threads = threads
--> 227 self._open_process()

File ~/Documents/dev/xopen/src/xopen/__init__.py:252, in _PipedCompressionProgram._open_process(self)
    250     stdout = self.fileobj  # type: ignore
    251 try:
--> 252     self.process = subprocess.Popen(
    253         self._program_args,
    254         stderr=self._stderr,
    255         stdout=stdout,
    256         stdin=subprocess.PIPE,
    257         close_fds=close_fds,
    258     )  # type: ignore
    259 except OSError:
    260     if self.closefd:

File ~/.pyenv/versions/3.12.9/lib/python3.12/subprocess.py:994, in Popen.__init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags, restore_signals, start_new_session, pass_fds, user, group, extra_groups, encoding, errors, text, umask, pipesize, process_group)
    975         raise ValueError(f"User ID cannot be negative, got {uid}")
    977 # Input and output objects. The general principle is like
    978 # this:
    979 #
   (...)    989 # are -1 when not using PIPEs. The child objects are -1
    990 # when not redirecting.
    992 (p2cread, p2cwrite,
    993  c2pread, c2pwrite,
--> 994  errread, errwrite) = self._get_handles(stdin, stdout, stderr)
    996 # From here on, raising exceptions may cause file descriptor leakage
    997 
    998 # We wrap OS handles *before* launching the child, otherwise a
    999 # quickly terminating child could make our fds unwrappable
   1000 # (see #8458).
   1002 if _mswindows:

File ~/.pyenv/versions/3.12.9/lib/python3.12/subprocess.py:1730, in Popen._get_handles(self, stdin, stdout, stderr)
   1727     c2pwrite = stdout
   1728 else:
   1729     # Assuming file-like object
-> 1730     c2pwrite = stdout.fileno()
   1732 if stderr is None:
   1733     pass

UnsupportedOperation: fileno
```

gnzsnz:
Updated reproduction steps, if I'm missing something, please let me know.

This time I'm going straight to `_PipedCompressionProgram` to test

```python
from io import BytesIO
import xopen

filelike = BytesIO()
msg = b"my test message"

# lz4
f = xopen._PipedCompressionProgram(
    filelike,mode="wb",program_settings=xopen._PROGRAM_SETTINGS["lz4"])

# gzip
f = xopen._PipedCompressionProgram(
    filelike,mode="wb",program_settings=xopen._PROGRAM_SETTINGS["gzip"])

# zstd
f = xopen._PipedCompressionProgram(
    filelike,mode="wb",program_settings=xopen._PROGRAM_SETTINGS["zstd"])
```

All these cases fail with `UnsupportedOperation: fileno`, full traceback already provided on my previous comment.

If these test are correct (I'm not 100% sure, but close to) then `lz4` is not introducing a new issue. This has been there but probably never detected.

Please let me know your comments. And how do you suggest moving forward. Personally, I believe that file-like is a great feature, but not a showstopper. But is your call.

marcelm:
Thanks for the updates!

Note #152, which I think made it possible to read from file-like objects (including `BytesIO`) when using `PipedCompressionProgram`, but AFAICS does not add support for *writing* to them.

The existing tests probably pass because they don’t end up using `_PipedCompressionProgram` but one of the in-process variants instead.

So yes, it seems that the problem is not in this PR.

I won’t have so much more time this week, but if you fix the tests (i.e., take out the failing ones using `BytesIO`), then this has my approval and maybe @rhpvorderman has time to merge if he approves as well.

gnzsnz:
updates:

- tests that require `_PipedCompressionProgram` skipped
- tests that require lz4 CLI on windows skipped (lz4 not installed on windows)

I believe we are good to merge this PR. please let me know your thoughts.

gnzsnz:
kind reminder

tests/test_xopen.py
```diff
@@ -690,6 +727,8 @@ def test_xopen_read_from_pipe(ext, threads):
 
 @pytest.mark.parametrize("threads", (0, 1))
 def test_xopen_write_to_pipe(threads, ext):
+    if ext == ".lz4" and shutil.which("lz4") is None:
+        pytest.skip("lz4 not installed")
```
rhpvorderman: These clauses will also skip when lz4 is not installed but python-lz4 is. That is not ideal.

src/xopen/__init__.py
```diff
@@ -551,6 +558,42 @@ def _open_zst(
...
+):
+    assert mode in ("rb", "ab", "wb")
+    if compresslevel is None:
+        compresslevel = XOPEN_DEFAULT_LZ4_COMPRESSION
+
+    if lz4 is not None and (mode == "rb" or (mode in ("ab", "wb") and threads == 0)):
+        # use Python bindings
+        f = lz4.frame.LZ4FrameFile(filename, mode, compression_level=compresslevel)
+        return f
+    else:
```
rhpvorderman: else clause is redundant. If the above case is true, the function returns.

src/xopen/__init__.py
```diff
@@ -683,6 +726,10 @@ def _detect_format_from_content(filename: FileOrPath) -> Optional[str]:
         elif bs[:4] == b"\x28\xb5\x2f\xfd":
             # https://datatracker.ietf.org/doc/html/rfc8478#section-3.1.1
             return "zst"
+        elif bs[:4] == b"\x04\x22\x4d\x18":
+            # https://en.wikipedia.org/wiki/LZ4_(compression_algorithm)
```
rhpvorderman: I'd rather have a link to a specification or other technical documentation here. There is no guarantee that wikipedia will keep listing the magic bytes.

tests/test_xopen.py
```diff
@@ -109,6 +113,8 @@ def test_binary(fname):
 def test_roundtrip(ext, tmp_path, threads, mode):
     if ext == ".zst" and threads == 0 and zstandard is None:
         return
+    if ext == ".lz4" and shutil.which("lz4") is None:
```
rhpvorderman: The availability of the lz4 program is already test on line 60. This clause is redundant, and skips testing even if python-lz4 is available.

rhpvorderman (changes requested):
Some minor nitpicks and one big gripe: skipping tests is not the same as fixing them. I'd suggest to remove all the added skipping code and properly investigate why some tests are failing.

src/xopen/__init__.py
gnzsnz: fixed on 4c18c24a98e100e6acefd48b832226c37b80aa0a

src/xopen/__init__.py
gnzsnz: fixed on 4c18c24a98e100e6acefd48b832226c37b80aa0a

tests/test_xopen.py
gnzsnz: the github action is testing across multiple OSes, which is 👍. but not all OS allow to install (that I'm aware) lz4/zstd. On linux and mac it's installed, but not on windows. which is making the "remaining" tests fail. the tests work the problem is that on the windows runner they fail.

this is the "fix" to account for windows runner.

tests/test_xopen.py
gnzsnz: line 60 is the fixture for extensions, '.lz4' is comming there, but this is needed, otherwise will fail when threads != 0. in this scenario we need lz4 CLI installed, and windows does not have it.

maybe

`if ext == ".lz4" and shutil.which("lz4") is None and threads != 0:`

src/xopen/__init__.py
```diff
@@ -551,6 +558,41 @@ def _open_zst(
...
+    if compresslevel is None:
+        compresslevel = XOPEN_DEFAULT_LZ4_COMPRESSION
+
+    if lz4 is not None and (mode == "rb" or (mode in ("ab", "wb") and threads == 0)):
+        # use Python bindings
+        f = lz4.frame.LZ4FrameFile(filename, mode, compression_level=compresslevel)
+        return f
+    # use CLI program
+    try:
+        return _PipedCompressionProgram(
```
marcelm, line 590: The idea here (and as discussed in earlier comments) was to first try to run the `lz4` binary with the `-T` option, but here, the `-T` option is not actually passed in because the `ProgramSettings` for `lz4` do not set a `threads_flag`. So both this and what is below in the `except` clause actually do the same thing.

src/xopen/__init__.py
```diff
@@ -551,6 +558,41 @@ def _open_zst(
...
+    except OSError:
+        _program_settings = _PROGRAM_SETTINGS["lz4"]
+        _program_settings.threads_flag = None
+        return _PipedCompressionProgram(
+            filename,
+            mode,
+            compresslevel,
+            threads,
+            program_settings=_program_settings,
+        )
```
marcelm, line 609: Also, this should fall back to using the bindings if calling the `lz4` binary fails.

tests/test_xopen.py
```diff
@@ -2,33 +2,37 @@
...
 except ImportError:
     zstandard = None
 
-
 # TODO this is duplicated in test_piped.py
 TEST_DIR = Path(__file__).parent
 CONTENT_LINES = ["Testing, testing ...\n", "The second line.\n"]
 CONTENT = "".join(CONTENT_LINES)
-extensions = ["", ".gz", ".bz2", ".xz"]
+extensions = ["", ".gz", ".bz2", ".xz", ".lz4"]
```
marcelm: `".lz4"` should not be added to `extensions` unconditionally (because it is possible that both the `lz4` bindings are missing *and* that the `lz4` binary is not installed). Instead it should be done as for `zstandard` below:
`if shutil.which("lz4") or lz4:` ...

This also makes it possible to remove most of the `if ext == ".lz4" and shutil.which("lz4") is None` lines below.

src/xopen/__init__.py
```diff
@@ -551,6 +558,41 @@ def _open_zst(
...
+        return _PipedCompressionProgram(
+            filename,
+            mode,
+            compresslevel,
+            threads,
+            program_settings=_PROGRAM_SETTINGS["lz4"],
+        )
+    except OSError:
+        _program_settings = _PROGRAM_SETTINGS["lz4"]
+        _program_settings.threads_flag = None
```
marcelm: This changes `_PROGRAM_SETTINGS["lz4"]` globally. Instead, only a local copy should be modified.

marcelm (commented):
I considered working on issue #172 but then realized this PR is still open and thought it would be fair to give it another close look. I found a couple of issues, so this is unfortunately not a quick touchup before it can be merged.

marcelm:
I pushed a couple of changes to address some of the things I noticed, but there are still failures. I removed most of the `pytest.skip` calls because these were just hiding problems. `xopen`ing an lz4 file should work unless *both* the lz4 binary is missing *and* the bindings are unavailable (which can only be the case on PyPy).

Dev-iL:
@gnzsnz @marcelm I'd like to help finalize this PR. Any objections to me rebasing this work into a new branch, fixing the build, and making a few small adjustments?

marcelm:
Fine with me.

PR #169, opened 2025-04-03
Use an SPDX license expression

marcelm:
This avoids a warning from setuptools, which will stop supporting the old style with "text =" next year.

See https://packaging.python.org/en/latest/specifications/license-expression/


marcelm:
Probably #170 needs to be merged first

PR #173, opened 2025-10-12
Drop support for Python 3.9

marcelm:
@rhpvorderman Do you know where the test failure on PyPy 3.10 comes from?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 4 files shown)


marcelm:
Ah, it is trying to build from the sdist because there are no PyPy wheels for isal 1.8.0.

marcelm:
It also seems we should bump to PyPy 3.11 as [PyPy 3.10 is end-of-life](https://cibuildwheel.pypa.io/en/stable/changelog/#v320).

rhpvorderman:
Sorry for being generally unresponsive. Recovery is going well now that we have finally root-caused a few things. Still only at one hour a day, and work projects do take priority.

PR #176, opened 2026-05-20
Drop support for Python 3.9; start supporting Python 3.14

msto:
Hey @marcelm  @rhpvorderman  - just a quick follow-up to #170, dropping support for 3.9 and adding 3.14 to the CI matrix

I opened a couple PRs upstream (https://github.com/pycompression/python-isal/pull/251, https://github.com/pycompression/python-zlib-ng/pull/79) to build the PyPy wheels for those dependencies. Probably cleanest to merge those in first, but you could also merge this as-is and add the PyPy support back after those are in. (If you go the latter route, it'd probably make sense to take `isal` and `zlib-ng` as CPython-only dependencies here before merge.)

This could easily be folded into #173 (and feel free to cherry-pick it in), just didn't want to step on an existing PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,9 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "pypy-3.9"]
+        # TODO: restore "pypy-3.11" once isal and zlib-ng ship PyPy wheels again
+        # (pycompression/python-isal#245, pycompression/python-zlib-ng#78).
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         optional-deps: [true]
         with-libs: [true]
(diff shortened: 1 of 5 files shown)


rhpvorderman:
Hi @msto, thanks for all your work. Indeed `isal` and `zlib-ng` should be tackled first. I will see what I can do. My time behind a computer is quite limited at this point due to health reasons. This might take a few weeks to see it through.

marcelm:
Hi and thanks @msto! I’m going to merge #177 instead with a couple of extra changes (I added you as co-author on the commit that adds 3.14 support).

BTW, do you need PyPy support? I’m currently totally fine with dropping it for good because it’s just a lot of extra work whenever we bump versions.

msto:
Thanks both!

@rhpvorderman sorry for what you're going through, wishing you the best.

@marcelm Merging #177 sounds good to me, I'll close this one. I do not need PyPy support, so no objections to dropping it.

PR #177, opened 2026-05-28
Modernize configuration; support 3.14; drop 3.9

marcelm:
This *drops PyPy from the test matrix* for now. Making PyPy work takes a lot of time and I don’t know whether a relevant number of people actually benefit from this, so this may become permanent.

Also:
* Migrate coverage configuration to `pyproject.toml`
* Migrate from `tox.ini` to `tox.toml`
* Switch from flake8 to ruff

Closes #173
Closes #176


.github/workflows/ci.yml
```diff
@@ -66,24 +66,24 @@ jobs:
     - name: Remove xz
       if: runner.os == 'Linux' && !matrix.optional-deps
       run: while which xz; do sudo rm $(which xz); done
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
```
msto, line 69: Small suggestion if you're interested - over at Fulcrum we've been getting into the habit of pinning to commit hashes rather than tags to promote reproducibility.

This blog post goes into more detail about why it's a useful security practice:

https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

marcelm: Interesting read! I agree this is a possible issue, but I consider a compromise of an essentially "built-in" action such as `actions/checkout` to be about as likely as GitHub’s CI infrastructure itself to be compromised, which is a risk I accept to live with. I admit the inconvenience of having to update long hashes and reduced readability weighs a bit stronger for me.

I’ll leave it as it is here for now, but will consider it for the projects where I use Dependabot.

rhpvorderman: Thanks @msto I agree with @marcelm for this project. Nevertheless interesting to consider for other projects.

PR #179, opened 2026-06-03
Rely on subprocess.Popen pipesize parameter rather than custom code for seting pipe size.

rhpvorderman:
As soon as I found the problem with pipes being 64KB by default and xopen issuing 128KB reads and writes I made a PR to CPython allowing to set the pipesize on subprocess.

Now we can finally pick the fruits of that labour!

As a side note, this is so useful. Forget bash. Pipe your programs together using python and set pipesize to 1MB. Enjoy much better speeds!

### DIFF
--- a/src/xopen/__init__.py
+++ b/src/xopen/__init__.py
@@ -17,7 +17,6 @@ import os
 import bz2
 import lzma
 import signal
-import pathlib
 import subprocess
 import tempfile
 import threading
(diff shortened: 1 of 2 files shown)


marcelm:
Awesome! (And wow, already six years since https://github.com/pycompression/xopen/pull/36#issue-681783076 ...)

